PR #70, opened 2015-11-10
Nav and nav title changes

androb:


androb:
@r12s - think you can merge this one?

PR #93, opened 2015-11-11
Fixed spellcheck typos

r12s:
Spell check should be one word, it now is.


androb:
@r12s as discussed let's go with "spell checker" as two words.

PR #241, opened 2015-11-30
replace code snippet double quote with single quote

r12s:


kilianc:
@r12s be aware that in HTML we should use double quote, so something like

```html
<script src="//cdn.tinymce.com/tinymce.min.js"></script>
```

should not be replaced by

```html
<script src='//cdn.tinymce.com/tinymce.min.js'></script>
```

kilianc:
Also since you are doing this, you may as well fix the 4 spaces indentation!

PR #259, opened 2015-12-04
Remove spaces after semicolon in block_formats configuration

ahoek:
If there are spaces after the semicolon in the block_formats configuration the translation doesn't work.


androb:
Thanks for the fix @ahoek! Merging now.

cc @r12s

PR #286, opened 2016-01-20
Fix syntax error

dok:


spocke:
Thanks for fixing this mistake.

PR #310, opened 2016-03-09
Add quotes to literal on JS example

izambl:
It throws an "red is not defined" error


spocke:
Thanks for fixing this one.

PR #316, opened 2016-03-17
Update importcss.md

mark-monteiro:
Use a custom no-op attribute (using vendor prefix notation) instead of `display:none` for the custom style start and end tags.


r12s:
imho `display:none` is suitable for this example. Marking closed.

btw Mark, I really like your website. Nice. Thanks for the contribution, always welcome.

PR #320, opened 2016-03-27
default value for image_caption is false instead of true

Cordemans:


r12s:
CLA not signed. Marking closed. Will investigate issue and open a PR if correct. Thanks for the heads up Cordemans!

PR #329, opened 2016-04-21
Fixed body class anchor

latenights:
Body class and body id example links to same anchor


r12s:
On the configuration pages, all instances of code example headings are H5 heads, producing multiple anchor instances with the same value on the page, which is not good (at all).

I'm closing this PR, not because it's not a good change, but because this issue should be fixed on all instances of #Example.

Thanks @latenights for noting this. I've created a task to action this.

PR #363, opened 2016-06-20
Fix types of code_dialog_width and height

mattgrande:
I noticed that `code_dialog_width` and `code_dialog_height` were both listed as `string`s, when they should be `number`s. This PR fixes that.


r12s:
thanks @mattgrande right you are! next time I'm in Ontario (or you're in the Bay Area) beers on TinyMCE :-)

PR #404, opened 2016-09-30
TINY-638: updated react integration docs to use create-react-app

fyrkant:


jayarjo:
What was the motivation behind the change, is `create-react-app` simpler to use or something?

fyrkant:
Yep, much simpler to use, it hides the webpack setup and all that (although you can eject if you want to and get all the configs to mess with). Feels like the perfect fit for something like this short guide, because it's supposed to be a guide about how to integrate TinyMCE and React, not about how to set up a webpack config and so on.

PR #415, opened 2016-10-12
Emit event in ng2 zone

kiddy2910:
We should emit events inside Angular zone to changes will be detected and affected.

Assume that this event call a method of Component used this directive, this method update a variable to a new value. If it doesn't run inside Angular zone, new value won't be detected and updated UI by Angular.

### DIFF
--- a/integrations/angular2.md
+++ b/integrations/angular2.md
@@ -81,7 +81,8 @@ import {
   AfterViewInit,
   EventEmitter,
   Input,
-  Output
+  Output,
+  NgZone
 } from '@angular/core';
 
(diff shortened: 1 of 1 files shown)


fyrkant:
Hi and thanks for the PR!

I'm not sure we need a zone.run here, because we are emitting the event to output Angular should be able to register the change by itself.

kiddy2910:
Because tinymce invoke its events outside from Angular zone, so if we don't call zone.run here, Angular cannot detect changes in binding model.

I took a lot of time to know that so I think it's maybe good for others :D

PR #427, opened 2016-11-03
fix #426, replace bower install --save-dev by --save

### LINKED ISSUE #426: A little issue with a command bower install
The file `https://github.com/tinymce/tinymce-docs/blob/master/integrations/angularjs.md` shows how to install `ui-tinymce` using bower:

    $ bower install angular-ui-tinymce --save-dev

But the option `--save-dev` is used for development dependencies, that is not the case here, because this is an angularjs module used to run tinymce in production environement.

So the command should be:

    $ bower install angular-ui-tinymce --save

rbecheras:
closes #426

### DIFF
--- a/integrations/angularjs.md
+++ b/integrations/angularjs.md
@@ -23,7 +23,7 @@ $ bower init
 We now install the angular-ui-tinymce package this will automatically install angular and tinymce.
 
 ```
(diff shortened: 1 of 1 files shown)


fyrkant:
Thank you!

rbecheras:
You're welcome!

PR #430, opened 2016-11-15
Changed the event type in the example of 'Blur' event listener

mwessner:
Obvious change in the provided example of 'Blur' event listener. It was faulty, probably due to c/p of the 'Focus' event listener example.


fyrkant:
Yep that seems to be a typo, thanks for the PR!

fyrkant:
Closes #431

PR #434, opened 2016-11-22
Angular2 directive implementation details with ngModel two-way binding

Abhijith-Nagaraja:
Currently, there are no examples for the following cases.

1. How to use Tinymce editor as a `Directive` instead of a `Component`
2. Two-way binding with ngModel.

This example shows both in one shot.

This is tested in latest anular2 version 2.2.1


fyrkant:
This looks great, but I think that it might be a little too complicated for this documentation and better suited as a standalone package on NPM. Then it would be easier for people to add to their own Angular 2 projects, easier to keep up to date and easier for users to file any issues they might run into. If you create such a package I would be happy to link it in the documentation!

Abhijith-Nagaraja:
I guess you are right. But currently, I am busy with other work and have no time to create a npm repo. For time being, for other's benefit, in the readme.md if you can point to my repo, then that would help others...

Something like this

## Sample directive implementation with ngModel two-way binding

You can check out the implementation here : [Directive Implementation with NgModel Binding](https://github.com/Abhijith-Nagaraja/tinymce-docs/blob/master/integrations/angular2.md#sample-directive-implementation-with-ngmodel-two-way-binding)

PR #447, opened 2016-12-06
Corrects typo

bdjnk:
"that causes and undo" -> "that cause an undo"

### DIFF
--- a/advanced/events.md
+++ b/advanced/events.md
@@ -272,7 +272,7 @@ tinymce.init({
 
 ### Change
 
-This event gets fired when changes are made inside the editor that causes and undo level to be added.
+This event gets fired when changes are made inside the editor that cause an undo level to be added.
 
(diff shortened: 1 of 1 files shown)


bdjnk:
As an aside, Ephox made this process unnecessarily aggravating.

androb:
@bdjnk we don't mean to! What went wrong?

bdjnk:
@androb The error message was cut off. The page it links to has too much text. I suggest: "Your email is private. Make it public so we can send you a legal document you must agree to. This is a requirement for contributing." The second error message was also cut off. It should begin: "We just sent you an email" Instead it links to the same unhelpful page, while the email is sent silently.

r12s:
Thank you for the commit @bdjnk and for point out the problem. Appreciated! I've created an issue in our tracker and we'll take a look. Most definitely want to improve this process. Thanks.

PR #464, opened 2016-12-28
Document compatible uglify-js usage

uncleyo:
Addresses https://github.com/tinymce/tinymce/issues/3412


advanced/usage-with-module-loaders.md
```diff
@@ -55,6 +55,22 @@ tinymce.init({
 });
 ```
 
+### Minification with UglifyJS2
```
fyrkant: This is great! 

But maybe put it at the end of the page, to better fit the flow of the other text? It feels like its about how to get started with TinyMCE with module loaders and then the minification in the middle of it, would probably fit better in the end I think!

uncleyo: Makes sense. Moved it to the bottom.

PR #498, opened 2017-03-02
Update documentation to reflect list plugin change

rmpressler:
Per https://github.com/tinymce/tinymce/issues/3342 `bullist` and `numlist` have been moved to list plugin but docs still show core


r12s:
Thanks @rmpressler greatly appreciate this update! I've also created an issue in my tracker to update the lists docs resulting from the change you highlighted.

PR #500, opened 2017-03-03
Update valid-elements.md

chasnz:
Changed synonymous to synonyms.


r12s:
@chasnz thanks again, "element synonymous" ain't a thing. Looks like I owe you beer.

PR #502, opened 2017-03-04
Update html example in first-steps.md

latvaaho:
This change was needed when taking a vanilla node.js with express project template to get form posting to work properly out-of-box. As I an not that experience front-end developer, it took a while to find out why I wasn't getting the editor's content into the body of submitted form. Finally the only issue was to have also name in addition to id defined for the textarea, and after that everything worked as a breeze.

### DIFF
--- a/get-started/first-steps.md
+++ b/get-started/first-steps.md
@@ -46,7 +46,7 @@ In this example, let's replace `<textarea id='mytextarea'>` with a TinyMCE edito
 <body>
 <h1>TinyMCE Quick Start Guide</h1>
   <form method="post">
-    <textarea id="mytextarea">Hello, World!</textarea>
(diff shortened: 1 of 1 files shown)


r12s:
@latvaaho thank you for the PR. I am not going to merge this for the following reason:

This should be a stand-alone example or integration document. Adding it to the Get Started docs may cause more confusion than it solves.

@latvaaho we do have an integrations page, https://www.tinymce.com/docs/integrations/ that might be a better place for this.

@fyrkant thoughts on whether this should be on integrations? Or perhaps the TinyMCE Community?

r12s:
No response from OP. Will open a new issue re a Node + Express integration. Marking closed.

PR #512, opened 2017-03-24
Replace the protocol-relative URL with https, and use it where possib…

gwils:
…le in the getting-started-cloud section

### DIFF
--- a/get-started-cloud/editor-plugin-version.md
+++ b/get-started-cloud/editor-plugin-version.md
@@ -13,14 +13,14 @@ When deploying the [TinyMCE editor via the Cloud]({{ site.baseurl }}/get-started
 Our documented default for loading TinyMCE via the Cloud is:
 
 ```js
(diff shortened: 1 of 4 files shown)


get-started-cloud/editor-plugin-version.md
```diff
@@ -13,14 +13,14 @@ When deploying the [TinyMCE editor via the Cloud]({{ site.baseurl }}/get-started
 Our documented default for loading TinyMCE via the Cloud is:
 
 ```js
-<script src="http://cloud.tinymce.com/stable/tinymce.min.js?apiKey=your_API_key"></script>
+<script src="{{ site.cdnurl }}?apiKey=your_API_key"></script>
```
techtangents: Why are you using a variable here, but not in the other URLs you've changed?

gwils: This URL uses `/stable/`, so we can substitute it for `site.cdnurl`, whereas most of the URLs on this page use a different midsection there, like `4` or `4.3`

PR #551, opened 2017-05-22
`textcolor_map` should be a list of strings, not a string itself

RealOrangeOne:
Not sure what the exact type definition should be
- `List`
- `Array`
- `Array<String>`
- `String[]`

Or, even better,

- `Object<String, String>`

### DIFF
--- a/plugins/textcolor.md
+++ b/plugins/textcolor.md
@@ -66,7 +66,7 @@ tinymce.init({
 
 This option allows you to specify a map of the text colors that will appear in the grid.
 
-**Type:** `List`
+**Type:** `Array`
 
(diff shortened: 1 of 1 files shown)


plugins/textcolor.md
```diff
@@ -66,7 +66,7 @@ tinymce.init({
 
 This option allows you to specify a map of the text colors that will appear in the grid.
 
-**Type:** `String`
+**Type:** `List`
```
fyrkant: Should probably be `Array`.

r12s:
@fyrkant can we get manual intervention from you to resolve this issue? thx

PR #575, opened 2017-08-14
add note about settings.special setting

alexpineda:


_includes/configuration/entity-encoding.md
```diff
@@ -4,6 +4,8 @@ This option controls how entities/characters get processed by TinyMCE. The value
 
 The base entities `<` `>` `&` `'` and `"` will always be entity encoded into their named equivalents. Though `'` and `"` will only be encoded within attribute values and `<` `>` will only be encoded within text nodes. This is correct according too the HTML and XML specs.
 
+TinyMCE will not encode any tags that are assigned to the special setting. By default this is settings.special = `script noscript style textarea`.
```
fyrkant: This is not correct, the defaults are `script noscript noframes noembed title style textarea xmp`, as seen here: https://github.com/tinymce/tinymce/blob/7d7a8db05d55bd5078849c23a342993588ce9995/src/core/src/main/js/html/Schema.js#L408

PR #601, opened 2017-09-25
Update angular2.md

clabough:
The existing implementation (i.e. only keyup) will not capture the final content if the last action the user takes is to use one of the toolbar options (e.g. bold, italics, etc.). Also capturing the change event will prevent this issue.

### DIFF
--- a/integrations/angular2.md
+++ b/integrations/angular2.md
@@ -107,6 +107,10 @@ export class SimpleTinyComponent implements AfterViewInit, OnDestroy {
       skin_url: 'assets/skins/lightgray',
       setup: editor => {
         this.editor = editor;
+        editor.on('Change', () => {
+          const content = editor.getContent();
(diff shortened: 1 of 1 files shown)


integrations/angular2.md
```diff
@@ -107,6 +107,10 @@ export class SimpleTinyComponent implements AfterViewInit, OnDestroy {
       skin_url: 'assets/skins/lightgray',
       setup: editor => {
         this.editor = editor;
+        editor.on('Change', () => {
```
fyrkant: You can instead just chain the event names, so just att `change` after `keyup` so it becomes `editor.on('keyup change', () => ...`.

fyrkant:
Closing this as we have released out own [@tinymce/tinymce-angular](https://github.com/tinymce/tinymce-angular) component and updated the documentation to use that instead.

PR #615, opened 2017-11-05
Changed tinymce-docs/integrations/angular2.md

hevilp:
elementId is never set. Setted in constructor


fyrkant:
It is set with the `@Input`.

hevilp:
Its declared as string yes, but setted?

fyrkant:
Yes it is set when you use the component:
```ts
<simple-tiny
  [elementId]="'my-editor-id'"
  (onEditorKeyup)="keyupHandlerFunction($event)"
  >
</simple-tiny>
```

PR #701, opened 2018-06-11
Resolve dependencies

shikhanansi:
Modify Gemfile.lock to resolve security vulnerability dependency caused by norogiri v1.8.1. Fix broken links in https://www.tinymce.com/docs/enterprise/check-spelling/custom/

Regards,
Shikha


shikhanansi:
Hi Dylan, 
I tried to fix the broken links without updating the nogori v1.8.1 to norogiri v1.8.2, but it only fixes the internal linking within the github repo. However, we still have some external documents like Textbox.io pointing to these links, which are still breaking if we don't update. 
Please review.
Regards,
Shikha

PR #703, opened 2018-06-18
Reflect correct default settings for autosave

gzzo:
The default value for `autosave_restore_when_empty` is `false`, according to:

https://github.com/tinymce/tinymce/blob/master/src/plugins/autosave/main/ts/api/Settings.ts#L29


shikhanansi:
@EkimChau could you please verify this change and provide feedback?
Regards,
Shikha

PR #717, opened 2018-07-02
Feature/DOC-15

shikhanansi:
https://ephocks.atlassian.net/browse/DOC-15. Added a section ## How to Enable the PowerPaste Plugin. Please review.

Regards,
Shikha

### DIFF
--- a/plugins/powerpaste.md
+++ b/plugins/powerpaste.md
@@ -11,7 +11,7 @@ The TinyMCE **PowerPaste** plugin automatically cleans up content from Microsoft
 
 ## How to Enable the PowerPaste Plugin
 
-To enable the PowerPaste plugin, the path to the PowerPaste folder has to be referenced from the [external plugins](../configure/integration-and-setup/#external_plugins) option in the `tinymce.init()`.
+To enable the **PowerPaste** plugin, the path to the PowerPaste folder has to be referenced from the [external plugins]({{site.baseurl}}/configure/integration-and-setup/#external_plugins) option in the `tinymce.init()`.
 
 Example:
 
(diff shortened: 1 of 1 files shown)


techtangents:
No issues from me. I just need someone else to review for technical accuracy.

techtangents:
No issues from my. I just need someone else to check the technical accuracy.

plugins/powerpaste.md
```diff
@@ -9,6 +9,21 @@ The TinyMCE **PowerPaste** plugin automatically cleans up content from Microsoft
...
+
+To enable the **PowerPaste** plugin, the path to the PowerPaste folder has to be referenced from the [external plugins]({{site.baseurl}}/configure/integration-and-setup/#external_plugins) option in the `tinymce.init()`.
+
+Example:
+
+```js
+tinymce.init({
+  selector: 'textarea',  // change this value according to your HTML
+  external_plugins: {
+    'powerpaste': 'http://www.testing.com/powerpaste.min.js',
```
TheSpyder: This doesn't seem right. Everywhere else in this file it's just `plugins: "powerpaste"` not `external_plugins`?

Most of this section is covered under `self-hosted installation` below. Perhaps it just needs to be improved if it's not obvious. And if these instructions need to be at the top, perhaps re-arrange so the two installation sections (cloud and self-hosted) are above `usage`?

TheRobQ: This should be specific only to importing PowerPaste with module loaders such as Webpack and Browserify (import/require).

TheSpyder: Fair enough. That isn't clear from the text, and would fit best as a sub-point under `self-hosted installation` :)

shikhanansi: Thank you for the review and feedback. Based on my understanding, I've made the following changes:
1. Moved the new section after the installation sections.
2. Moved the Usage section after the installations and new module loader section.
3. Changed the new section to accommodate feedback.
Please review and let me know if you have any feedback.

shikhanansi:
I agree with @TheSpyder. Usage should come before installation. I have moved Usage before installations. Merging this request since there are no more concerns with the accuracy.
Kind regards,
Shikha

PR #733, opened 2018-08-14
fixed local api reference and updated readme.md

spocke:
This makes it easier to generate api docs from local files. Updated the readme so that the apollo team can generate docs locally.

### DIFF
--- a/_scripts/api-reference-local.sh
+++ b/_scripts/api-reference-local.sh
@@ -2,13 +2,19 @@
 
 set -e
 
+if [[ $1 -eq 0 ]] ; then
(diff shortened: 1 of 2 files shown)


_scripts/api-reference-local.sh
```diff
@@ -2,13 +2,19 @@
 
 set -e
 
+if [[ $1 -eq 0 ]] ; then
+  echo 'You need to specify the root tinymce directory to generate the source for'
+  echo './_scripts/api-reference-local ../tinymce'
```
EkimChau: './_scripts/api-reference-local.sh ../tinymce'

PR #753, opened 2018-09-20
Remove redundant note

sevanspowell:
- This note asks the user to append `https://` to the site url, but the site url
already contains `https://`.


sevanspowell:
@shikhanansi I've looked through the documentation source and haven't found anywhere else where this is a problem.

To re-iterate the problem, this note asks the user to prepend `https:` to the url, but the url variable that we're using when building the docs is `https://cloud.tinymce.com/stable/tinymce.min.js`, so it already has `https:`.

What should my next step be? Should I mirror this discussion in Jira?

PR #801, opened 2018-11-01
updated docs for updated wordcount plugin

fyrkant:
Updated wordcount page and removed info about old settings, also added keyboard shortcut.


mhideo:
This plugin adds word count functionality to TinyMCE by placing a counter on the right edge of the status bar. A toolbar button opens a dialog box with both word and character counts. Clicking the wordcount in the statusbar switches between counting words and characters.

lnewson:
Looks good to me, thanks @fyrkant! Need a review before we can merge though, so added @shikhanansi as a reviewer :)

PR #815, opened 2018-11-21
TINY-2253: Updated the docs to match changes to the UI elements menu item references

lnewson:
When specifying menu items for context menus, menu buttons and sub menus, they can now be represented by a space separated list of menu items (the same as in the init config) or alternatively as an array of menu items. This differs from the previous behaviour which was to only allow an array of menu items.

If it helps, here's an example from the table menu button showing using the space separated string format:
```
  editor.ui.registry.addMenuButton('table', {
    tooltip: 'Table',
    icon: 'table',
    fetch: (callback) => callback('inserttable tableprops deletetable | cell row column')
  });
```

Note: This hasn't been released yet, so we maybe should hold off merging for now.

### DIFF
--- a/_includes/codepens/contextmenu-section/index.js
+++ b/_includes/codepens/contextmenu-section/index.js
@@ -13,10 +13,16 @@ tinymce.PluginManager.add('my-example-plugin', function (editor) {
       return !element.src ? [] : ['image'];
     }
   });
+
+  editor.ui.registry.addContextMenu('link', {
+    update: (element) => {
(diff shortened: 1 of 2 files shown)


lnewson:
@shikhanansi just as a heads up I revert the codepen examples largely after talking to Andy, just to try and keep them as simple as possible.

lnewson:
@shikhanansi this has been merged now, so not sure if you want to merge this just yet? or wait until the next release happens.

PR #820, opened 2018-11-22
TINY-2230: Updated codepen examples for the new nested menu items type

lnewson:
We currently don't have any docs I could find that document the different menu items (which matches our discussion today). So for now I've just updated any thing that currently exists, which appears to only be the codepen examples.

Note: This JIRA is still on review, so we should hold off merging for now.


lnewson:
@shikhanansi this has been merged now, so not sure if you want to merge this just yet? or wait until the next release happens.

PR #830, opened 2018-11-29
Convert inlite plugin to Quick Toolbar plugin

androb:
This needs review, and should not be merged until the code changes are made.


lnewson:
I'm not sure which is correct but the code looks to be using `quickbars` for the name and settings prefix, while this is using `quickbar`. So we probably need to fix either the code or this PR I guess?

lnewson:
Asked around and from what I understand the decision was it should be the plural form (ie `quickbars`), because there are multiple bars provided. So in that case can we get this updated to reflect that please?

To be clear, the code is using the following:

Plugin name: quickbars
Settings:
- quickbars_insert_toolbar
- quickbars_selection_toolbar

plugins/quickbars.md
```diff
@@ -0,0 +1,118 @@
...
+
+##### Example enabling both context toolbars with their default controls:
+
+```js
+tinymce.init({
+    selector: "div.tinymce",
+    plugins: [ 'quickbars' ],
+    toolbar: false,
+    menubar: false,
+    quickbars: true
```
lnewson: Should this have been changed from `inline: true` to `quickbars: true`? I suspect this may have accidentally been swapped since `inline` is very similar to `inlite`.

plugins/quickbars.md
```diff
@@ -0,0 +1,118 @@
...
+```
+#### Example disabling the Quick Insert toolbar:
+
+```js
+tinymce.init({
+    selector: "div.tinymce",
+    plugins: [ 'quickbars' ],
+    toolbar: false,
+    menubar: false,
+    quickbars: true,
```
lnewson: Should this have been changed from `inline: true` to `quickbars: true`?

plugins/quickbars.md
```diff
@@ -0,0 +1,118 @@
...
+
+#### Example disabling the Quick Selection minibar:
+
+```js
+tinymce.init({
+    selector: "div.tinymce",
+    plugins: [ 'quickbars' ],
+    toolbar: false,
+    menubar: false,
+    quickbars: true,
```
lnewson: Should this have been changed from `inline: true` to `quickbars: true`?

plugins/quickbars.md
```diff
@@ -0,0 +1,118 @@
...
+
+##### Example:
+
+```js
+tinymce.init({
+    selector: "div.tinymce",
+    plugins: [ 'quickbars' ],
+    toolbar: false,
+    menubar: false,
+    quickbars: true,
```
lnewson: Should this have been changed from `inline: true` to `quickbars: true`?

plugins/quickbars.md
```diff
@@ -0,0 +1,118 @@
...
```
lnewson: Should this have been changed from `inline: true` to `quickbars: true`?

plugins/quickbars.md
```diff
@@ -0,0 +1,118 @@
...
```
lnewson: Should this have been changed from `inline: true` to `quickbars: true`?

plugins/quickbars.md
```diff
@@ -0,0 +1,118 @@
...
+```js
+tinymce.init({
+    selector: "div.tinymce",
+    plugins: [ 'quickbars' ],
+    toolbar: false,
+    menubar: false,
+    inline: true
+});
+```
+#### Example disabling the Quick Insert toolbar:
```
TheSpyder: This didn't make the release 🤦‍♂️ we're rushing a fix in now
see https://github.com/tinymce/tinymce/issues/4626

PR #833, opened 2018-11-29
TINY-1903: added docs on custom roots for drive

spocke:
Updated docs for custom root paths and image max size for drive.


shikhanansi:
@spocke Can we merge this branch into develop for the Beta release tomorrow?

PR #840, opened 2018-12-03
Update TinyMCE codepen url for beta

lnewson:
Updated the tinymce url for the codepens, to use the new cloud service for tinymce 5


lnewson:
@idooo think I may have got ahead of myself here, so feel free to discard this if needed.

lnewson:
Thanks Alex for the catch on using https :+1:

PR #844, opened 2018-12-05
added page embed plugin documentation page

fyrkant:


fyrkant:
I added a comment on x-frame-options header that can block some pages from embedding. 

Also, I don't think "create" is the correct word in "Creating a Page Embed toolbar button", it should probably be "add" as in "Adding a Page Embed toolbar button"?

PR #858, opened 2018-12-20
TINY-2994: expanded the format options docs

spocke:
This expands the docs on the various options for formats. Since I'm going away on vacation for about 6 weeks I won't be able to modify this PR much if it contains grammar issues etc. So I hope you @shikhanansi can massage this PR if needed.

### DIFF
--- a/_includes/configuration/formats.md
+++ b/_includes/configuration/formats.md
@@ -8,7 +8,7 @@ Check out the [custom formats example]({{ site.baseurl }}/demo/format-custom/) f
 
 Formats are objects where the value is either an object with format options or an array of format variants. 
 
-When a format is applied the first item, the formats array will be applied. However, when trying to match the format to remove or update the UI, the other formats in the array is considered as well. So the first format is a kind of primary format, and the rest is secondary formats.
+When a format is applied the first item, the formats array will be applied. However, when trying to match the format to remove or update the UI, the other formats in the array is considered as well. So the first format is a kind of primary format, and the rest are secondary formats.
 
 The following is an example of an array of format variants:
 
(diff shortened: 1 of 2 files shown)


spocke:
This might also need backporting but if you are to make changes to this then I suggest that you @shikhanansi does that backporting to 4.x.

_includes/configuration/formats.md
```diff
@@ -42,18 +44,320 @@ Some built-in formats `fontsize`, `fontname`, `forecolor`, `hilitecolor` uses a
...
+    ]
+  }
+});
+```
+
+
+
+#### deep
+
+Enables you to control if formats of children with the same matching format should be removed as well. This is set to true by default on selector formats since if you select a table
```
lnewson: @spocke should this instead say "is set to false by default" (and the same below)? As looking at https://github.com/tinymce/tinymce/blob/master/src/core/main/ts/fmt/FormatRegistry.ts#L29 it mentions it's false by default and doing a quick test by registering a format it was false when using a selector.

_includes/configuration/formats.md
```diff
@@ -1,20 +1,32 @@
...
+For example, a format is a style that gets applied to text when the bold button is enabled inside the editor. TinyMCE is equipped with a text formatting engine that allows specifying what it should produce when the user clicks (in this example) the bold button.
 
 Check out the [custom formats example]({{ site.baseurl }}/demo/format-custom/) for a demonstration of this option.
 
-Formats is an object where the values is either an object with format options or an array of format variants. When applying a format the first item in this formats array will be applied but when matching the format when removing or updating the ui the other formats in the array is considered as well. So the first format is a kind of primary format and the rest is secondary formats.
+Formats are objects where the value is either an object with format options or an array of format variants. 
+
+When a format is applied the first item, the formats array will be applied. However, when trying to match the format to remove or update the UI, the other formats in the array is considered as well. So the first format is a kind of primary format, and the rest is secondary formats.
```
EkimChau: - the rest is secondary formats

+ the rest are secondary formats

shikhanansi: Fixed

PR #866, opened 2019-01-08
TINY-2848: Updated dialog checkbox codepen examples to match checkbox state changes

lnewson:
Previously checkboxes were represented as a string of either "checked", "unchecked" or "intermediate", however it was decided to make them just have two states and make them use a boolean (true/false) instead of a string.

I couldn't find any mention of the data types returned for each UI component, so there should only be these examples that need updating.


lnewson:
Thanks Shikha! :+1:

PR #876, opened 2019-01-15
DOCINBOX-12: Added some more details about dialog migration from 4.x -> 5.x

lnewson:
Reworded the onpostrender -> onSetup details, to clarify that's for toolbar and menu components. Also expanded on some of the removed and new settings for dialogs.

@shikhanansi I think this should cover what we discussed for this, but more than happy to discuss and make any additional changes.

### DIFF
--- a/migration-from-4x.md
+++ b/migration-from-4x.md
@@ -331,8 +331,9 @@ Dialogs are still opened via the `editor.window.open(spec)` api, however a numbe
 | ------------------- | --------------- |
 | height | The dialog component now uses CSS3 and a predefined `small`, `medium`, and `large` template to specify the dimensions. |
 | width | The dialog component now uses CSS3 and a predefined `small`, `medium`, and `large` template to specify the dimensions. |
-| bodyType | The `bodyType` has been merged into the `body` setting.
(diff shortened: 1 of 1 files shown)


migration-from-4x.md
```diff
@@ -323,7 +323,25 @@ editor.ui.registry.addMenuItem('example', {
...
+
+##### Removed settings:
+
+| **Removed setting** | **Description** |
+| ------------------- | --------------- |
+| height | The dialog component now uses CSS3 and a predefined `small`, `medium`, and `large` template to specify the dimensions. |
+| width | The dialog component now uses CSS3 and a predefined `small`, `medium`, and `large` template to specify the dimensions. |
+| bodyType | `bodyType` has been merged into the `body` setting.
+| onpostrender | The dialog configuration instead now includes an `initialData` setting for providing the initial state and an api to fetch or update the data.
+| url | The `iframe` dialog component has been removed, which makes the `url` setting irrelevant.
```
lnewson: Wasn't sure `irrelevant` was the right phrase here. I wondered if instead it should have maybe been `obsolete`?

PR #880, opened 2019-01-16
DOCINBOX-12: Added details about onChange being merged into a single callback

lnewson:
@shikhanansi this is the part I mentioned I missed during standup. `onCancel` and `onTabChange` are new callbacks and previously changes were detected by an `onchange` callback on each component. However now there's just a single callback function with the change data.

If it helps, here's some examples (in case you want to reword or elaborate on what I added):

TinyMCE 4.x:
```
const config = {
  title: 'Insert Link',
  body: [
    {
      name: 'text',
      type: 'textbox',
      size: 40,
      label: 'Text to display',
      onchange () {
        data.text = this.value();
      }
    }
  ]
};
```

TinyMCE 5.0:
```
const config = {
  title: 'Insert Link',
  body: {
    type: 'panel',
    items: [
      {
        name: 'text',
        type: 'input',
        label: 'Text to display'
      }
    ]
  },
  onChange (api, changeData) {
    if (changeData.name === 'text') {
      // Do something with the text to display changes
    }
  }
};
```


lnewson:
Looks perfect, thanks Shikha!

PR #881, opened 2019-01-16
TINY-2907: Added codepen example for context forms/toolbars and updated nav links

lnewson:


lnewson:
@shikhanansi I'll add the context form example we mentioned this morning as well, so you might want to ignore this for the moment :)

lnewson:
@shikhanansi this should be ready to review now, as I've added the context form example as well. I should mention I changed a couple example heights, as they were quite large and left a lot of whitespace at the bottom of the page.

lnewson:
Also adding @EkimChau as a reviewer, just to make sure it matches what we want for examples.

PR #892, opened 2019-01-18
Fix to Cat Dog Machine codepen

martyf:
initialData for isdog is expecting a string of “checked”, “unchecked” or “indeterminate”, not boolean.

Fix changes this for the initialData, as well as the onSubmit logic to correctly select “cat” or “dog”

Signed-off-by: Marty Friedel <marty@mity.com.au>

### DIFF
--- a/_includes/codepens/dialog-pet-machine/index.js
+++ b/_includes/codepens/dialog-pet-machine/index.js
@@ -31,11 +31,11 @@ var dialogConfig =  {
   ],
   initialData: {
     catdata: 'initial Cat',
-    isdog: false
+    isdog: 'unchecked'
   },
   onSubmit: function (api) {
     var data = api.getData();
-    var pet = data.isdog ? 'dog' : 'cat';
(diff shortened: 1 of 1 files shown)


shikhanansi:
@EkimChau please review this PR

lnewson:
This is actually changing in our next release and checkboxes will be switching to using a boolean as part of our API review before the stable release. If you test with the latest version from git, you should see it requires a boolean instead of a string. It seems the docs change for that, might have got pushed out a little earlier sorry. 

For reference: https://github.com/tinymce/tinymce/issues/4758

shikhanansi:
@martyf  thank you for your contribution. Our engineers have received your suggestion and are working on providing a solution. This issue is fixed in the upcoming release of TinyMCE. I will provide a status update once we've released the latest version.
Your inputs are valuable to us. Thank you for helping Tiny deliver a better user experience.
Kind regards,
Shikha

TheSpyder:
This should've been closed at the time :)

PR #795, opened 2018-10-24
#4643: Add visualchars_default_state option for visualchars plugin

mat3e:
`visualchars_default_state` for `visualchars` plugin inspired by `visualblocks_default_state` from `visualblocks`

PR #896, opened 2019-01-21
Re-add visualchars_default_state option for visualchars plugin

lnewson:
This reinstates the original version (https://github.com/tinymce/tinymce-docs/pull/795) that got reverted in commit 9af454417b7e361c168fb0dfe5f491f5ed05c617 due to it being merged ahead of time.


lnewson:
@shikhanansi the related PR has been merged now, so this should be okay to add back into the docs now (or at least it should be merged before GA).

PR #899, opened 2019-01-22
TINY-3174: Fixed issues with Tiny Drive demos

lnewson:
The drive demo was using an old version of TinyMCE, so moved it to being a codepen, so it'll use the same  TinyMCE version as the rest of the demos. Also fixed an issue where the images would 404 in the drive plugin demo, because the image paths needed to move up one more level.


lnewson:
Found a couple other minor issues sorry, so fixed them up as well.

lnewson:
Thanks all! :+1:

PR #903, opened 2019-01-22
Removed file_browser_callback documentation

TheSpyder:
This setting has been removed in v5. I also added a note to the migration guide.


_includes/configuration/file-picker-callback.md
```diff
@@ -1,6 +1,6 @@
 ## file_picker_callback
 
-This hook can be used to add custom file picker to those dialogs that have it. Internally we support this in *Image*, *Media* and *Link* dialogs. Option replaces deprecated [file_browser_callback](#file_browser_callback). New hook provides a way to update values of other fields in the dialog.
+This hook can be used to add custom file picker to those dialogs that have it. Internally we support this in *Image*, *Media* and *Link* dialogs. This replaces the [`file_browser_callback`](https://www.tiny.cloud/docs/configure/file-image-upload/#file_browser_callback) (removed in version 5.0) option. The new `file_picker_callback` provides a way to update values of other fields in the dialog.
```
TheSpyder: This link likely won't work once v5 docs are published, which is why I left it out...

shikhanansi: I've confirmed with DX. The link will change. However, it will be easier to redirect them later than reviewing the docs and adding the relevant ones.

TheSpyder: I figured no link would be better than a broken link, but ok.

PR #911, opened 2019-01-31
Fixed up issues found in the content formatting examples

lnewson:
Was reviewing these docs, as I wanted to link them for another issue. However I noticed that a number of the examples aren't being formatted as examples, so this should fix that.


lnewson:
Here's an example of what it currently looks like: https://www.tiny.cloud/docs-beta/configure/content-formatting/#selector

PR #928, opened 2019-02-10
DOC-182: Documented the autocompleter api

lnewson:
I should note that the autocompleter defaults in 5.0.0 are a little different, as we've just adjusted them a little different. For example the default `minChars` in 5.0.0 was 0 instead of 1 and the default `matches` predicate was just to always return true, instead of it matching at the start of a word. So I've targeted this more so at the 5.0.1 release.

As always, please feel free to change anything if you want :smiley:

### DIFF
--- a/ui-components/autocompleter.md
+++ b/ui-components/autocompleter.md
@@ -46,45 +46,9 @@ The `fetch` function is called whenever the trigger `char` is pressed and the `m
 | ---- | ----- | ----------- |
 | hide | () => void | Hides the autocompleter menu. |
 
-### Example of an autocompleter that completes an email address domain:
-
-```js
-tinymce.init({
-  selector: '#editor',
-  setup: (editor) => {
-    editor.ui.registry.addAutocompleter('myCustomAutocompleter', {
-      ch: '@',
-      columns: 1,
-      matches: (rng) => rng.startOffset !== 0,
-      fetch: (text) => {
-        return new Promise((resolve) => {
-          resolve([
-            {
-              value: '@example.com',
-              text: 'example.com',
-              icon: 'user'
-            },
-            {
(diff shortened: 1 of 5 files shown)


ui-components/autocompleter.md
```diff
@@ -0,0 +1,90 @@
...
+layout: default
+title: Autocompleter
+title_nav: Autocompleter
+description: Add a custom autocompleter to TinyMCE 5.0.
+keywords: autcomplete
+---
+
+## Overview and use case
+
+An `autocompleter` enables users to quickly find and select from a pre-populated list of values as they type, leveraging searching and filtering. It provides suggestions to insert while the user is typing into the content. For example, typing **:amp** should show the ampersand item in the menu. Pressing `esc` should close the autocomplete menu.
```
blackcathikari: Add an example from the emoji plugin?

ui-components/autocompleter.md
```diff
@@ -0,0 +1,90 @@
...
+The two arguments this method take are:
+
+* `name` - A unique name for the autocompleter.
+* `configuration` - An object containing the user's configuration.
+
+### Configuration options
+
+| Name | Value | Requirement | Description |
+| ---- | ----- | ----------- | ----------- |
+| ch | string | Required | The character to trigger the autocompleter. |
```
blackcathikari: We should really rename this to something more descriptive in the code. Now would be a good time, before we merge these docs. @TheSpyder @EkimChau thoughts?

ui-components/autocompleter.md
lnewson: Technically charmap, but yeah that'd be a good idea to add that :+1:

lnewson: Done :smile:

PR #934, opened 2019-02-14
add content_css setting to all examples

fyrkant:
We recently had a customer confused on how this plugin works, not understanding that the importcss plugin only uses the css brought in by the content_css setting so I thought it would be better if the content_css setting is set on all of the examples.


shikhanansi:
@fyrkant can we merge this PR now?

fyrkant:
Yes go ahead.

PR #936, opened 2019-02-18
DOC-112: Menu item docs and examples

lnewson:
This takes @blackcathikari's initial draft, makes a couple tweaks (eg ordering the menu items alphabetically) and also adds an example to the demo page for this.

@shikhanansi I also changed the path to the custom menu item demo, as it's not related to toolbars. So we'll also likely need a redirect setup at some stage for `/docs/demo/custom-toolbar-menu-item/` to `/docs/demo/custom-menu-item/`


migration-from-4x.md
```diff
@@ -193,10 +193,10 @@ The following new methods have been added for creating and using new components:
 | editor.ui.registry.addContextForm: (name, spec) | [Context form]({{site.baseurl}}/ui-components/contextform/) |
 | editor.ui.registry.addContextMenu: (name, spec) | [Context menu]({{site.baseurl}}/ui-components/contextmenu/) |
 | editor.ui.registry.addMenuButton: (name, spec) | [Menu Button]({{site.baseurl}}/ui-components/typesoftoolbarbuttons/#menubutton) |
-| editor.ui.registry.addNestedMenuItem: (name, spec) | [Nested Menu Item]({{site.baseurl}}/migration-from-4x/#custommenuitems) |
+| editor.ui.registry.addNestedMenuItem: (name, spec) | [Nested menu item]({{site.baseurl}}/ui-components/menuitems/#nestedmenuitems) |
 | editor.ui.registry.addSplitButton: (name, spec) | [Split Button]({{site.baseurl}}/ui-components/typesoftoolbarbuttons/#splitbutton) |
 | editor.ui.registry.addToggleButton: (name, spec) | [Toggle Button]({{site.baseurl}}/ui-components/typesoftoolbarbuttons/#togglebutton) |
-| editor.ui.registry.addToggleMenuItem: (name, spec) | [Toggle menu item]({{site.baseurl}}/migration-from-4x/#custommenuitems) |
+| editor.ui.registry.addToggleMenuItem: (name, spec) | [Toggle menu item]({{site.baseurl}}/ui-components/menuitems/#togglemenuitems) |
```
blackcathikari: So we're doing alphabetical rather than grouping by type?

blackcathikari:
On the demo page's path change, the V4 demo really shouldn't be called "Custom Toolbar Menu Item" either. It should just be "Custom Menu Item" in both. I assume Lee's change would need to be ported to somewhere else to fix that?

ui-components/menuitems.md
```diff
@@ -0,0 +1,124 @@
...
+
+* Create a shortcut for an action or a series of actions that the user repeatedly does.
+* Create a button for custom behavior.
+
+## How to create custom menu items
+
+The methods for adding custom menu items are in the UI Registry part of the editor API editor.ui.registry. The API has three methods for adding menu items:
+
+* `editor.ui.registry.addMenuItem`(identifier, configuration)
+* `editor.ui.registry.addToggleMenuItem`(identifier, configuration)
```
blackcathikari: Let's order these alphabetically too? Might as well be consistent.

ui-components/menuitems.md
```diff
@@ -0,0 +1,124 @@
...
+| onAction | (api) => void | required | Function invoked when the menu item is clicked. |
+
+#### API
+
+| Name | Value | Description |
+| ---- | ----- | ----------- |
+| isDisabled | () => boolean | Checks if the menu item is disabled. |
+| setDisabled | (state: boolean) => void | Sets the menu item's disabled state. |
+
+### Nested menu items
```
blackcathikari: I wrote this very quickly, hence there aren't any examples for the nested or toggle menu item sections. We should really add some, even if they're very simple. Could even take the `editor.ui.registry.add...` examples from the demo page and throw them in here?

migration-from-4x.md
lnewson: No idea, but a good question. I'd have to defer to @shikhanansi here, as it was already sorted alphabetically.

ui-components/menuitems.md
lnewson: Agreed! I completely missed this one, so thanks for the catch :)

ui-components/menuitems.md
lnewson: :+1: I largely reused them, though made them use alerts instead of inserting content.

ui-components/menuitems.md
lnewson: Fixed :smiley:

shikhanansi:
@lnewson please merge this PR.

PR #955, opened 2019-02-26
Feature/tiny 3143

EkimChau:


changelog.md
```diff
@@ -8,7 +8,32 @@ class: changelog
 
 {% capture changelog %}
 
+## Version 5.0.1 February 21, 2019
```
androb: @shikhanansi normally it would be best to start the changelog entry with what has been added, then go onto what has been fixed... and end with what has been removed.

lnewson: That'd be my fault sorry, as atm the process/doc for changelogs is to just copy what we have in the TinyMCE changelog.txt file basically. I'll make a note of that and be sure to reformat it in that order when submitting the changelog for the docs.

androb: Cool thanks @lnewson  - we probably need to keep an eye on grammar, capitalization etc. It's not a BIG deal... but just adds to the professionalism :)

TheSpyder: normally a team lead or equivalent should review the changelog before release, that didn't happen for 5.0.1

PR #956, opened 2019-02-26
TINY-3143: fix local paths, should not impact prod, only dev

EkimChau:


shikhanansi:
@EkimChau can we merge this PR now?

PR #989, opened 2019-03-14
Feature/docinbox 20

shikhanansi:
@Grade-Zero please review. The Case Change plugin isn't available on /5 yet. So the plugin will not load until it is released.


_includes/codepens/casechange/index.js
```diff
@@ -0,0 +1,6 @@
+tinymce.init({
+  selector: "textarea",
```
lnewson: This should be `"textarea#casechange"`, otherwise it could try and override other areas on the page. We maybe should also consider adding this to the menu since it's not in the default menus, otherwise you'll only be able to access this via the toolbar button. eg:

```
  ...
  toolbar: 'casechange',
  menubar: 'tools',
  menu: {
    tools: { title: 'Tools', items: 'casechange' }
  }
  ...
```

_includes/codepens/checklist/index.js
```diff
@@ -0,0 +1,6 @@
+tinymce.init({
+  selector: "textarea",
```
lnewson: This should be "textarea#checklist", otherwise it could try and override other areas on the page.

demo/checklist.md
```diff
@@ -0,0 +1,15 @@
...
+title_nav: Checklist
+description: Add checklists to the content.
+keywords: lists todo checklist
+controls: toolbar button
+---
+
+
+## Live example
+
+This example shows how to use the **Checklist** plugin to add interactive checklists to your content. For more information on the Checklist plugin, see the [docs]().
```
lnewson: Should this have a link to the docs?

_includes/codepens/casechange/index.js
shikhanansi: Fixed

demo/checklist.md
shikhanansi: Fixed

PR #1031, opened 2019-04-30
feature/DOC-305: Added information on how to create a sidebar toggle button in the toolbar

wonderbugtiny:
`Custom sidebar` section within `UI components`:
- Added a line within `Editor sidebar API` heading about how to create a sidebar toggle button in the toolbar.
- Added more information to the `icon` sub-heading.
- Replaced the `settings` icon with `comment` icon in the example code.

### DIFF
--- a/ui-components/customsidebar.md
+++ b/ui-components/customsidebar.md
@@ -15,7 +15,7 @@ The sidebar API allows developers to add sidebars on editor instances in a simil
 
 This is the syntax for the addSidebar function: `editor.ui.registry.addSidebar(name:String, spec:Object)`
 
-Including the sidebar name in the `toolbar` option will create a sidebar toggle button with the icon specified in the `onSetup` function.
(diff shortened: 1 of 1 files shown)


ui-components/customsidebar.md
```diff
@@ -15,6 +15,8 @@ The sidebar API allows developers to add sidebars on editor instances in a simil
 
 This is the syntax for the addSidebar function: `editor.ui.registry.addSidebar(name:String, spec:Object)`
 
+Including the sidebar name in the `toolbar` option will create a sidebar toggle button with the icon specified in the `onSetup` function.
```
lnewson: This isn't quite correct, as the toolbar icon is already created when just registering the sidebar.

lnewson: Maybe instead we could say something like this?
```
When a new sidebar is registered, a corresponding toolbar button is also created using the same name. This button can then be included in the toolbar by adding the name to the [`toolbar`](({{site.baseurl}}/configure/editor-appearance/#toolbar) configuration option.
```

wonderbugtiny: Thank you. It is now fixed.

PR #1059, opened 2019-06-03
GH-1057 Fix buttons types and update URL dialog page

### LINKED ISSUE #1057: URL Dialog "Button Types" incorrectly lists "close"
The [documentation](https://www.tiny.cloud/docs/ui-components/urldialog/#buttontypes) states that "close" is a valid button type.

However this results in an error, and the error says to use either "cancel" or "custom".

Changing to "cancel" works as I would expect "close" to work - the documentation doesn't really specify when or why you would want or need to use "cancel" vs "close"

blackcathikari:
Fix #1057 and update URL dialogs page to match the new dialogs page. @lnewson please review. I tried not to touch the messaging section, and just updated other parts.

### DIFF
--- a/ui-components/urldialog.md
+++ b/ui-components/urldialog.md
@@ -7,26 +7,32 @@ keywords: dialog urldialog api
 ---
 ## Overview
 
-A URL dialog is a special TinyMCE UI component which loads an external web page inside a dialog (sometimes referred to as `modals`). This differs from the regular dialogs that use supported components to render an interactive dialog inside the application. URL dialogs are useful for very complex use cases, where the supported dialog components cannot be used — for example, a custom file manager that is loaded inside a TinyMCE dialog.
+A URL dialog is a special TinyMCE UI component which loads an external web page inside a dialog (sometimes referred to as `modals`). URL dialogs are useful for very complex use cases, where the supported components for TinyMCE's standard dialogs cannot be used. For example, a custom file manager that is loaded inside a TinyMCE dialog would probably require a URL dialog.
 
-The most basic configuration structure is:
+> Note: [Standard TinyMCE dialogs]({{site.baseurl}}/ui-components/dialog/) should suffice for most use cases, and may be simpler to configure.
+
+### Basic example
+
+The configuration for a basic URL dialog might look like this:
 
 ```js
-const urlDialogConfig = {
+tinymce.activeEditor.windowManager.openUrl({
    title: 'Just a title',
    url: 'http://www.tiny.cloud/example.html'
-}
+});
 ```
 
 ## URL dialog configuration
 
-A URL Dialog configuration has two main parts:
+A URL Dialog configuration has three main parts to match the three main parts of the dialog's UI:
 
-* **Title** This is the title of a dialog.
+* **Title:** The title of the dialog. This will display in the header of the dialog.
 
-* **URL** The external pages URL to load inside the dialog.
+* **URL:** The URL of the external page to load inside the dialog.
 
-### Config options
+* **Buttons: optional -** An array of [footer buttons](#footerbuttons) that are displayed in the dialog's footer.
+
+### Configuration options
 
 | Name    | Value  | Requirement | Description |
 | ------- | ------ | ----------- | ----------- |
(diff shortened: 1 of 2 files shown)


ui-components/urldialog.md
```diff
@@ -7,91 +7,90 @@ keywords: dialog urldialog api
...
 | width   | number | optional    | The width of the dialog in pixels. |
 | height  | number | optional    | The height of the dialog in pixels. |
-| buttons | UrlDialogButton[] | optional | An optional array of button configurations to render in the footer. See [Footer components](#footercomponents) configuration. |
-| onAction | (api) => void | optional | Function invoked when a custom button is clicked. |
-| onCancel | (api) => void | optional | Function invoked when the dialog is cancelled. |
-| onClose | () => void | optional | Function invoked when the dialog has been closed. |
-| onMessage | (api, data) => void | optional | Function invoked when a message is received from the external page. |
-
-### Footer components
+| buttons | FooterButton[] | optional | An optional array of [footer buttons](#footerbuttons) to render in the footer of the dialog. |
```
lnewson: It feels like we're repeating `footer` too much here maybe? Maybe instead change the first reference to `button configurations` or even just simply `buttons`?

ui-components/urldialog.md
```diff
@@ -7,91 +7,90 @@ keywords: dialog urldialog api
...
 
-**Name:** The name property on the button is used as an ID attribute to identify the dialog component. For example, when `name: foobutton` is defined and a user clicks on that button, the dialog `onAction()` handler will fire and provide an object containing the name of the dialog component, e.g., `details.name = 'foobutton'`. This will allow developers to create a click handler for **foobutton**.
+#### Configuration
 
-**Text:** This will be the text displayed on the button. For example, `text: ‘do magic’` will create a button with text **do magic**.
+| Name | Type | Requirement | Description |
+| ---- | ---- | ----------- | ----------- |
+| type | `'button'` | required | The component type. Must be `'button'`. |
+| text | string | required | Text to display in the button **if `icon` is not specified**. Also used for the button's `title` attribute. |
```
lnewson: The bold markup around `if icon is not specified` hides that icon is a keyword and I'm not sure it makes sense to have that much bold? Maybe just make `if` bold or remove it altogether?

ui-components/urldialog.md
```diff
@@ -7,91 +7,90 @@ keywords: dialog urldialog api
...
+| name | string | optional | A identifier for the button. If not specified, the button will be assigned a randomly generated `name`.  |
```
lnewson: This should probably be `An identifier`.

ui-components/urldialog.md
```diff
@@ -7,91 +7,90 @@ keywords: dialog urldialog api
...
+| icon | string | optional | Name of the icon to be displayed. Must correspond to an icon in the icon pack. **When configured, the button will display the icon instead of text.** |
+| primary | boolean | optional | Whether to style the button as a primary or secondary button. |
```
lnewson: Maybe add the default here, like the line below, as we used to include that?

ui-components/urldialog.md
```diff
@@ -7,91 +7,90 @@ keywords: dialog urldialog api
...
+> Note: Buttons do not support mixing icons and text at the moment.
 
-**Disabled:** (Value: Boolean; Default: `False`): When set to `true`, the button will be disabled when the dialog loads.
+#### Button types and event callbacks
 
-**Primary:** (Default: `False`): When set to `true`, the button will be colored to stand out. The color will depend on the chosen [skin]({{site.baseurl}}/general-configuration-guide/customize-ui/#skins).
+The different footer button types will invoke different callbacks when clicked:
 
-**Align:** (Default: 'end'): This will define the position of the button in the footer. When set to `end`, the button will be positioned on the right side of the dialog. When set to `start`, the button will be positioned on the left side of the dialog.
+* A **Cancel** type button will invoke the `onCancel` and `onClose` callback functions provided in the dialog configuration. These callback functions are also fired when a user clicks the `X` button in the top right of the dialog. **TinyMCE assumes that a dialog will not have multiple Cancel buttons.**
```
lnewson: `TinyMCE assumes that a dialog will not have multiple Cancel buttons.` is confusing and misleading to me. I'd suggest just removing it or rewording it since two buttons are possible, it's just that they'd always have the same effect (so it wouldn't make much sense to include two).

PR #1061, opened 2019-06-03
Td 251

spocke:
Docs for TinyDrive 1.3 mostly the standalone/plugin api. Did some changes to the codepen code though since it was needed in order to support both tinymce and tinydrive api endpoints.


_includes/codepens/drive-standalone-upload/example.js
```diff
@@ -0,0 +1,16 @@
+var button = document.querySelector('.tinydrive-standalone-demo-upload');
+var percent = document.querySelector('.tinydrive-standalone-demo-percent');
+
+button.addEventListener('click', function () {
+  tinydrive.upload({
+    token_provider: '/your-local/jwt-provider',
+    path: '/hello',
+    name: 'hello.txt',
+    blob: new Blob('Hello world!'),
```
lnewson: This doesn't look like it'd work, as it'd need to be `new Blob(['Hello world!']),` I believe?

_config.yml
```diff
@@ -5,7 +5,8 @@ baseurl: ""
 shared_baseurl: ""
 syntax_highlight_theme: "tomorrow-night"
 cdnurl: "https://cloud.tinymce.com/5/tinymce.min.js"
-codepen_url: "https://cloud.tinymce.com/5/tinymce.min.js?apiKey=qagffr3pkuv17a8on1afax661irst1hbr4e6tbv888sz91jc"
+tinymce_codepen_url: "https://cloud.tinymce.com/5/tinymce.min.js?apiKey=qagffr3pkuv17a8on1afax661irst1hbr4e6tbv888sz91jc&tinydrive=dev"
+tinydrive_codepen_url: "https://cdn.tiny.cloud/1/qagffr3pkuv17a8on1afax661irst1hbr4e6tbv888sz91jc/tinydrive/dev/tinydrive.min.js"
```
lnewson: Should this be using `dev`? As that's the v4 branch atm. Theoretically that should be fine but given this is meant to be the v5 docs I'm wondering if it should use `5-dev` instead?

Note: I'm assuming we should also update this to `5-stable` once it's actually released.

tinydrive/tinydrive-api/standalone.md
```diff
@@ -0,0 +1,282 @@
+---
+layout: default
+title: Tiny Drive Standalone API
+title_nav: Standalone API
+description: Description of the Tiny Drive standalone api
+keywords: tinydrive api
+---
+
+Tiny Drive can be used as a generic file manager separate from tinymce this is referred to as standalone mode.
```
lnewson: `TinyMCE` maybe?

tinydrive/tinydrive-api/standalone.md
```diff
@@ -0,0 +1,282 @@
...
+tinydrive.pick({
+  dropbox_app_key: '<your dropbox app key>'
+}).then((result) => {
+  console.log(result.files);
+});
+```
+
+### `filetypes`
+
+This setting enables to restrict what types of files you want do display based on [file type]({{site.baseurl}}/tinydrive/introduction/#filetypes) categories. For example if your app needs to insert images only then you can specify `['image']` in the file types array.
```
lnewson: Maybe instead of `enables to restriction what types`, use `enables restricting what types`? and similarly the same in plugin.md.

tinydrive/tinydrive-api/standalone.md
```diff
@@ -0,0 +1,282 @@
...
+tinydrive.pick({
+  max_image_dimension: 1024
+}).then((result) => {
+  console.log(result.files);
+});
+```
+
+### `target`
+
+This setting enables you to render tinydrive within a target element by using a css selector.
```
lnewson: `Tiny Drive` maybe, since that seems to be what's used elsewhere?

lnewson:
Just as a side note, we probably should avoid merging this until R9 has been released as well (delayed until tomorrow), since we'd need to merge develop into master.

lnewson:
@spocke feel free to merge this whenever you want to now, as R9 docs have been released 😃

PR #1074, opened 2019-06-05
feature/DOC-324: Updated toolbar group labels spec and note for using Toolbar drawer with multiple toolbars

wonderbugtiny:
Updated toolbar spec to include details on adding toolbar group labels and a note on using toolbar drawer with multiple toolbars

### DIFF
--- a/_includes/configuration/toolbar-drawer.md
+++ b/_includes/configuration/toolbar-drawer.md
@@ -6,6 +6,8 @@ Use the [toolbar]({{site.baseurl}}/configure/editor-appearance/#toolbar) option 
 
 **Type: String**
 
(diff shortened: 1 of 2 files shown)


_includes/configuration/toolbar-drawer.md
```diff
@@ -6,6 +6,8 @@ Use the [toolbar]({{site.baseurl}}/configure/editor-appearance/#toolbar) option
 
 **Type: String**
 
+> **Note:** The toolbar drawer will not be applicable to [multiple toolbars]({{site.baseurl}}/configure/editor-appearance/#usingmultipletoolbars) or when using the [toolbar(n)]({{site.baseurl}}/configure/editor-appearance/#toolbarn) option.
```
lnewson: `is not available when using` maybe instead of `will not be applicable to`?

lnewson:
As discussed last week, we should delay merging this until after the 5.0.7/R10 docs have been released.

- [ ] Don't merge until after R10 docs have been released.

PR #1080, opened 2019-06-10
TINY-3658: Added docs for new imagetools_fetch_image function

lnewson:
Added this as I noticed it was missing when writing up the internal R10 release notes for Bart. @spocke feel free to change anything if I've got it wrong or if you want to word things differently.

### DIFF
--- a/plugins/imagetools.md
+++ b/plugins/imagetools.md
@@ -88,7 +88,7 @@ tinymce.init({
 
 ### `imagetools_fetch_image`
 
-This option can be used to define a custom fetch function, which provides another way to access images in complex situations. The function will be passed the HTML element of the image to be fetched and should return a `Blob` containing the content.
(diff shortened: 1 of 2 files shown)


plugins/imagetools.md
```diff
@@ -86,6 +86,27 @@ tinymce.init({
...
+**Type:** `Function`
+
+##### Example
+
+```js
+tinymce.init({
+  selector: "textarea",  // change this value according to your HTML
+  toolbar: "image",
+  plugins: "image imagetools",
+  imagetools_fetch_image: function (img) {
```
spocke: It should return a promise you can change this to say return Promise.resolve(new Blob(...)) maybe.

lnewson: I opted against `Promise.resolve` as it's unlikely someone would actually end up using that and instead used the more async method. So that should hopefully be a more realistic example.

PR #1098, opened 2019-06-19
Some small capitalization suggestions

androb:
Added some basic edits


lnewson:
@androb just double checking is this meant to go to master or develop? Just asking so that if it is intended to be a hotfix, then we can also make sure we merge it back into develop afterwards 😀

androb:
What would you prefer Lee? It isn't urgent. I probably should have just
selected Develop.

lnewson:
It doesn't bother me either way, so I guess let's leave it as is? as there's no reason this can't go directly to master 👍

PR #1108, opened 2019-07-01
TINY-3774: Added 5.0.9 release notes

lnewson:
Little bit late sorry, but here's the release notes for the S11 release. I wasn't exactly too sure how to handle the fact that we skipped 5.0.8 so I just included a note at the top of the release notes.


release-notes/release-notes509.md
```diff
@@ -0,0 +1,74 @@
...
+keywords: releasenotes newfeatures deleted technologypreview bugfixes knownissues
+---
+
+## Overview
+
+The Release Notes provide high-level coverage of the improvements and additions that have been implemented in TinyMCE 5.0.9 and document known problems in this release, as well as important bug fixes, deprecated functionality, and other details.
+
+TinyMCE 5.0.9 release adds improvements to the TinyMCE editor to improve the overall user experience.
+
+> **Note:** TinyMCE 5.0.8 community was not released to enterprise due to blockers found during QA. As such TinyMCE 5.0.9 enterprise contains changes from both TinyMCE 5.0.8 and 5.0.9 community releases.
```
blackcathikari: `TinyMCE 5.0.8 community was not released to enterprise due to blockers found during QA` sounds... not quite formal enough to me? Or rather, I'm not sure if `blockers` is general terminology. Maybe `...due to issues found during QA.` instead?

release-notes/release-notes509.md
```diff
@@ -0,0 +1,74 @@
...
+
+The **Media** plugin now supports the `M4A` audio format. Inserting a m4a file will now be embedded using a HTML5 `<audio>` element.
+
+For more information on the media plugin refer to the [documentation]({{site.baseurl}}/plugins/media/).
+
+## Updates and enhancements
+
+### TinyMCE
+
+The new TinyMCE 5.0.9 editor comes with significant changes to the previous versions. Improved solutions have been introduced for unnecessary and obsolete configuration options. A comprehensive list of the updates and fixed issues is available in the [changelog]({{site.baseurl}}/changelog/#version509june262019).
```
blackcathikari: ...This paragraph is getting rather like a broken record. We could perhaps just use the last sentence with a few words added? The rest isn't super necessary.

lnewson: Yeah, I feel the same way, but it's been there for every version atm so I'm hesitant to remove it tbh. Maybe we can remove the middle sentence, as I think that's the worse part and really doesn't make sense to me?

PR #1121, opened 2019-07-12
TINY-3890: Added 5.0.11 release notes

lnewson:


lnewson:
Declining this, as we won't be releasing 5.0.11 now.

PR #1127, opened 2019-07-17
WEBSITE-1703 - Cart/checkout buttons only visible when needed

Grade-Zero:
In place for both desktop and mobile
Checked in all browsers


_includes/js/store.js
```diff
@@ -0,0 +1,42 @@
...
+      credentials: "include",
+      method: "GET",
+      mode: "cors"
+    }).then(function (response) {
+      return response.json()
+    }).then(function (response) {
+      if (response.itemCount > 0) {
+        showStoreButtons();
+      }
+    });
```
glennpenridge: Probably want to put a `catch()` here so we can stop errors from breaking anything.

_includes/template/bodycontents.html
```diff
@@ -214,10 +214,13 @@ <h2 class="light">{{ page.description }}</h2>
 {% include_remote {{ site.origin }}/_docs/footer.html css="body > *" %}
 <script src="{{ site.shared_baseurl }}{{ site.baseurl }}/scripts/common.min.js"></script>
 <script src="{{ site.shared_baseurl }}{{ site.baseurl }}/scripts/docs.min.js"></script>
+<script src=https://cdn.jsdelivr.net/npm/promise-polyfill@8.1/dist/polyfill.min.js></script>
+<script src=https://cdn.jsdelivr.net/npm/whatwg-fetch@3.0/dist/fetch.umd.min.js></script>
```
glennpenridge: Should we be host these files ourselves?

PR #1128, opened 2019-07-17
TINY-3054: Added new dialog details from changes done for search and replace

lnewson:
Note: I'm not 100% sure these are right, as they are just what I was able to find that had been added as part of the accessibility checker/search and replace dialog changes.

### DIFF
--- a/ui-components/dialog.md
+++ b/ui-components/dialog.md
@@ -75,7 +75,7 @@ tinymce.activeEditor.windowManager.open({
 | buttons | FooterButton[] | required | An array of [footer buttons](#footerbuttons) to render in the footer of the dialog. |
 | size | `'normal'`, `'medium'` or `'large'` | optional | default: `normal` - Dialog size options. |
 | initialData | object | optional | An object containing initial values for the dialog's panel components. |
-| <a class="anchor" id="onactionfunction"></a>onAction | `(dialogApi, details) => void` | optional | Function invoked when a user interacts with a `button` type panel component, clicks a **Custom** type footer button, or clicks an item in a **Menu** type footer button. |
+| onAction | `(dialogApi, details) => void` | optional | Function invoked when a user interacts with a `button` type panel component, clicks a **Custom** type footer button, or clicks an item in a **Menu** type footer button. |
 | onSubmit | `(dialogApi) => void` | optional | Function invoked when a **Submit** type footer button is clicked. |
 | onCancel | `(dialogApi) => void` | optional | Function invoked when the dialog is cancelled. The dialog header's close button and a **Cancel** type footer button invoke this function. |
 | onChange | `(dialogApi, details) => void` | optional | Function invoked when the value of an `input` type panel component changes. |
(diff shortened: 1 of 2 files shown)


ui-components/dialog.md
```diff
@@ -136,15 +136,15 @@ The Help plugin's dialog is an example of a tab panel dialog.
 
 ### Footer buttons
 
-A **button** is a clickable component that can contain text or an icon. There are two types of buttons (primary and secondary buttons), though the only difference is that they are styled differently. Primary buttons are intended to stand out. The color will depend on the chosen [skin]({{site.baseurl}}/general-configuration-guide/customize-ui/#skins).
+A **button** is a clickable component that can contain text or an icon. There are three types of buttons (primary, secondary and menu buttons). Primary and secondary buttons will perform an action when clicked, however they are styled differently. Primary buttons are intended to stand out. The color will depend on the chosen [skin]({{site.baseurl}}/general-configuration-guide/customize-ui/#skins). Menu buttons instead of performing an action will open a menu with more options when clicked.
```
blackcathikari: "Menu buttons instead of performing an action will open a menu with more options when clicked." -> "Menu buttons will open a menu with more options when clicked, instead of performing an action."?

TheSpyder:
Maybe we should let this dialog get through QA and other reviews before we publish the API changes and lock it down forever 🤔

lnewson:
+1 more than happy to let this sit until we want it to be released. I just didn't want the new configuration/functionality to be completely forgotten.

lnewson:
This is currently blocked, awaiting changes done in TINY-3952

ui-components/dialog.md
```diff
@@ -184,6 +183,66 @@ See the [dialog configuration options](#configurationoptions) documentation for
 }
 ```
 
+##### Dialog menu buttons
+
+A dialog menu button is a split button that can be used to provide a drop-down list of items in a dialog footer.
```
lnewson: Split button isn't the right term here, it should be a `dropdown button` or something instead, as split buttons are one that has two parts to it: one that is the primary and one that opens a sub menu.

ui-components/dialog.md
```diff
@@ -184,6 +183,66 @@ See the [dialog configuration options](#configurationoptions) documentation for
...
+
+The following options can be specified for a dialog menu button:
+
+| Name    | Value                  | Requirement | Description                                                                    |
+| ------- | ---------------------- | ----------- | ------------------------------------------------------------------------------ |
+| items   | array                  | required    | An array of [dialog menu items](#dialogmenuitems).                             |
+| text    | string                 | optional    | Text to display if no icon is found.                                           |
+| icon    | string                 | optional    | Name of the icon to be displayed. Must correspond to an icon in the icon pack. |
+| tooltip | string                 | optional    | Text for button tooltip.                                                       |
+| onSetup | (api) => (api) => void | optional    | default: () => () => {} - Function that's invoked when the button is rendered. |
```
lnewson: Asked and this actually shouldn't be exposed sorry! So we shouldn't include it in the docs and we'll remove it in the source.

ui-components/dialog.md
blackcathikari: Agreed. This sentence also seems... awkward. I think mostly the usage of "provide" seems odd...

ui-components/dialog.md
```diff
@@ -183,6 +183,66 @@ See the [dialog configuration options](#configurationoptions) documentation for
...
+
+###### Dialog menu items
+
+The following options can be specified for a dialog menu button _item_:
+
+| Name     | Value   | Requirement | Description                                                                                                                                           |     |
+| -------- | ------- | ----------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- | --- |
+| name     | string  | required    | Identifier for the dialog menu item which is passed to the [_dialog `onAction` callback_](#onactionfunction).                                         |     |
```
lnewson: This is also used to set the initial data/state, in case you want to add that.

ui-components/dialog.md
tylerkelly13: Deleted

ui-components/dialog.md
tylerkelly13: added

> `name` can be used with [initialData](#configurationoptions) to set the initial state.

ui-components/dialog.md
```diff
@@ -165,11 +165,11 @@ The different footer button types will invoke different callbacks when clicked:
 * A **Submit** type button will invoke the `onSubmit` callback function provided in the dialog configuration.
 * A **Cancel** type button will invoke the `onCancel` and `onClose` callback functions. These callback functions are also fired when a user clicks the `X` button in the top right of the dialog.
 * A **Custom** type button will invoke the `onAction` callback function, and pass it the button's `name` in the `details` object. This allows developers to create a click handler for each **Custom** type footer button in the dialog. See the [Redial example](#interactiveexampleusingredial) for an example of how to use this.
+* A **Menu** type button will fetch a list of options and display them in a drop-down menu. For details, see: [Dialog menu buttons](#dialogmenubuttons).
```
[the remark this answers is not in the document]
tylerkelly13: Menu ?items? invoke onAction, and pass the `name` of the menu item that's clicked.?

PR #1144, opened 2019-08-05
updated browser support page to use ticks rather than 'X's

tylerkelly13:
updated browser support page to use ticks (&#10003;) rather than 'X's to improve UX


tylerkelly13:
Whole page needs work, deferring these changes at this time.

PR #1158, opened 2019-08-08
MENTIONS-17: Added new autocompleter API details

lnewson:
Note: I haven't included the mentions specific changes in this, as it might be released separately to 5.0.14.

### DIFF
--- a/ui-components/autocompleter.md
+++ b/ui-components/autocompleter.md
@@ -33,7 +33,7 @@ The two arguments this method take are:
 | maxResults | number | Optional | default: 10 - The maximum number of results that should be fetched. |
 | minChars | number | Optional | default: 1 - The minimum number of characters that must be typed before the autocompleter will trigger (excluding the trigger char). |
 
-The `fetch` function is called whenever the trigger `char` is pressed and the `matches` predicate returns `true`. It is a function that takes the matched text pattern and returns a promise containing matching results. This allows for asynchronous fetching of the autocompleter items. The `fetchOptions` passed to the `fetch` function is by default an empty object, however using the [reload](#api) API additional arguments can be passed to fetch a different set of results. The results should be a list of objects with the following details:
+The `fetch` function is called whenever the trigger `char` is pressed and the `matches` predicate returns `true`. It is a function that takes the matched text pattern and returns a promise containing matching results. This allows for asynchronous fetching of the autocompleter items. The `fetchOptions` passed to the `fetch` function is by default an empty object, however using the [reload](#api) API additional options can be passed to fetch a different set of results. The results should be a list of objects with the following details:
 * `value`: Value of the item. This will be passed to the `onAction` callback when selected.
 * `text`: Text to display for the item.
 * `icon`: Name of the icon to be displayed. Must be a single unicode character or correspond to an icon in the icon pack.
--- a/ui-components/autocompleter.md
+++ b/ui-components/autocompleter.md
@@ -26,7 +26,7 @@ The two arguments this method take are:
 | Name | Value | Requirement | Description |
 | ---- | ----- | ----------- | ----------- |
 | ch | string | Required | The character to trigger the autocompleter. |
-| fetch |  (pattern: string, maxResults: number) => Promise<AutocompleterItem[]> | Required | A function that is passed the current matched text pattern and the maximum number of expected results. The function should return a promise containing matching results. |
+| fetch |  (pattern: string, maxResults: number, fetchOptions: Record<string, any>) => Promise<AutocompleterItem[]> | Required | A function that is passed the current matched text pattern and the maximum number of expected results. The function should return a promise containing matching results. |
 | onAction | (api, rng: Range, value: string) => void | Required | A function invoked when a fetched item is selected. |
 | columns | number or 'auto' | Optional | default: auto - The number of columns to show. If set to `1` column, then icons and text are displayed, otherwise only icons are displayed. |
 | matches | (rng: Range, text: string, pattern: string) => boolean | Optional | default: isStartOfWord - A predicate function that takes a range, the current text node content and the matched text content and returns a boolean indicating if the autocompleter should trigger. |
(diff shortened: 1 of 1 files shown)


ui-components/autocompleter.md
```diff
@@ -26,7 +26,7 @@ The two arguments this method take are:
 | Name | Value | Requirement | Description |
 | ---- | ----- | ----------- | ----------- |
 | ch | string | Required | The character to trigger the autocompleter. |
-| fetch |  (pattern: string, maxResults: number) => Promise<AutocompleterItem[]> | Required | A function that is passed the current matched text pattern and the maximum number of expected results. The function should return a promise containing matching results. |
+| fetch |  (pattern: string, maxResults: number, fetchOptions: Record<string, any>) => Promise<AutocompleterItem[]> | Required | A function that is passed the current matched text pattern and the maximum number of expected results. The function should return a promise containing matching results. |
```
TheSpyder: Probably best to note `fetchOptions` is optional 🤔

lnewson: It's not optional? it's defaulted and would just be an empty object. However yeah we probably should clarify it'll be the value passed from reload, otherwise will be an empty object.

TheSpyder: Oh right sorry I was looking at it backwards

yeah, defaulted seems good

lnewson: I did some re-organizing and added a sentence about the defaults and how fetchOptions is passed from reload, so let me know what you think :)

ui-components/autocompleter.md
```diff
@@ -26,14 +26,14 @@ The two arguments this method take are:
...
 | maxResults | number | Optional | default: 10 - The maximum number of results that should be fetched. |
 | minChars | number | Optional | default: 1 - The minimum number of characters that must be typed before the autocompleter will trigger (excluding the trigger char). |
 
-The `fetch` function is called whenever the trigger `char` is pressed and the `matches` predicate returns `true`. It is a function that takes the matched text pattern and returns a promise containing matching results. This allows for asynchronous fetching of the autocompleter items. The results should be a list of objects with the following details:
+The `fetch` function is called whenever the trigger `char` is pressed and the `matches` predicate returns `true`. It is a function that takes the matched text pattern and returns a promise containing matching results. This allows for asynchronous fetching of the autocompleter items. The `fetchOptions` passed to the `fetch` function is by default an empty object, however using the [reload](#api) API additional arguments can be passed to fetch a different set of results. The results should be a list of objects with the following details:
```
TheSpyder: singular additional argument, it's an object but there is only one ;)

lnewson: Oops, that was meant to be `options`, not `arguments`. Thanks for the catch Andy!

ui-components/autocompleter.md
```diff
@@ -26,14 +26,14 @@ The two arguments this method take are:
 | Name | Value | Requirement | Description |
 | ---- | ----- | ----------- | ----------- |
 | ch | string | Required | The character to trigger the autocompleter. |
-| fetch |  (pattern: string, maxResults: number) => Promise<AutocompleterItem[]> | Required | A function that is passed the current matched text pattern and the maximum number of expected results. The function should return a promise containing matching results. |
+| fetch |  (pattern: string, maxResults: number, fetchOptions: Record<string, any>) => Promise<AutocompleterItem[]> | Required | A function that is passed the current matched text pattern, the maximum number of expected results and any additional fetch options. The function should return a promise containing matching results. |
```
blackcathikari: Reading the following paragraphs, is fetchOptions optional? Or are users meant to always pass something, even if it's an empty object? I'm not quite clear on that.

We also don't have a standard way of marking optional args from memory though so not sure how we'd even do that...

ui-components/autocompleter.md
```diff
@@ -26,14 +26,14 @@ The two arguments this method take are:
...
 | maxResults | number | Optional | default: 10 - The maximum number of results that should be fetched. |
 | minChars | number | Optional | default: 1 - The minimum number of characters that must be typed before the autocompleter will trigger (excluding the trigger char). |
 
-The `fetch` function is called whenever the trigger `char` is pressed and the `matches` predicate returns `true`. It is a function that takes the matched text pattern and returns a promise containing matching results. This allows for asynchronous fetching of the autocompleter items. The results should be a list of objects with the following details:
+The `fetch` function is called whenever the trigger `char` is pressed and the `matches` predicate returns `true`. It is a function that takes the matched text pattern and returns a promise containing matching results. This allows for asynchronous fetching of the autocompleter items. The `fetchOptions` passed to the `fetch` function is by default an empty object, however using the [reload](#api) API additional options can be passed to fetch a different set of results. The results should be a list of objects with the following details:
```
blackcathikari: I think the last sentence - "The results..." - should be a new paragraph? On first read I thought that was referring just to the reload() section.

ui-components/autocompleter.md
```diff
@@ -45,6 +45,7 @@ The `fetch` function is called whenever the trigger `char` is pressed and the `m
 | Name | Value | Description |
 | ---- | ----- | ----------- |
 | hide | () => void | Hides the autocompleter menu. |
+| reload| (fetchOptions: Record<string, any>) => void | Hides the autocompleter menu and fetches new menu items. The additional `fetchOptions` will be passed to the `fetch` callback. |
```
blackcathikari: Re: my first comment, is fetchOptions optional here too, or needs to be defaulted? Also, I think "additional" isn't necessary?

ui-components/autocompleter.md
blackcathikari: Actually, I'm also confused on what/where/why fetchOptions is used. Should we explain that more? Is it just meant to be any data that is needed by the fetch implementation to return the right data?

lnewson: Users never call `fetch` it's a function you define that gets called by the system.

lnewson: It's also described in more detail below, but if you have any better suggestions by all means feel free to make a change.

ui-components/autocompleter.md
lnewson: Removed `additional` as you're right it's not needed. Also it's not optional as per the types, as if a dev is reloading then you'd want to get different data so you'd want to pass different fetch options.

ui-components/autocompleter.md
lnewson: Fixed

PR #1169, opened 2019-08-19
Add missing settings to Image plugin settings

tylerkelly13:


lnewson:
I checked the source and the anchor settings only apply to `file` url input types, but the image plugin uses an `image` url input, so the anchor settings don't apply to the image plugin.

PR #1174, opened 2019-08-23
5.0.14 release notes, initial upload

tylerkelly13:


release-notes/release-notes5014.md
```diff
@@ -0,0 +1,56 @@
...
+
+![Redesigned Accessibility Checker Dialog]({{site.baseurl}}/images/accessibility_checker.png)
+
+> **Note:** We don't currently support any additional accessibility rules, but it does enable variations of the current rules to be checked. For example, enabling `aaa` will check the color contrast is 7:1 instead of 4.5:1.
+
+## Updates and enhancements
+
+### PowerPaste 5.0.2
+
+PowerPaste 5.0.2 now provides a way for developers to keep unsupported image source URLs in the HTML, instead of PowerPaste completely removing them. To enable set the new `powerpaste_keep_unsupported_src` setting to `true`, which will store the original `src` in the `data-image-src` attribute. This allows developers to replace the images manually using post-processing. For example, if the src is a `file://` URL and they have access to the local filesystem, they can add the image during post-processing of the posted content.
```
lnewson: `s/To enable set/To enable, set/`?

release-notes/release-notes5014.md
```diff
@@ -0,0 +1,56 @@
...
+Two new settings have been introduced to allow developers to set the Accessibility Level and HTML version they wish to comply to. The new settings are:
+
+* `a11ychecker_level`  - a, aa or aaa (aa by default)
+* `a11ychecker_html_version` - html4 or html5 (html4 by default)
+
+Accessibility Checker 2.1.0 provides a new streamlined design for the accessibility checker dialog:
+
+![Redesigned Accessibility Checker Dialog]({{site.baseurl}}/images/accessibility_checker.png)
+
+> **Note:** We don't currently support any additional accessibility rules, but it does enable variations of the current rules to be checked. For example, enabling `aaa` will check the color contrast is 7:1 instead of 4.5:1.
```
lnewson: Maybe reword this, as I don't think we want it talking about "We"? Maybe something like this?

> Currently no additional accessibility rules are supported, however variations of the current rules can now be checked against different levels. For example, enabling `aaa` will check the color contrast is 7:1 instead of 4.5:1.

release-notes/release-notes5014.md
lnewson: Also we probably need to remove "they" in the last sentence and replace it with something like a reference to the developer?

tylerkelly13: replacing with 'you'

release-notes/release-notes5014.md
tylerkelly13: Reworded
> Currently, additional accessibility rules are not supported, but variations of the current rules can be checked.

release-notes/release-notes5014.md
lnewson: I thought we were supposed to avoid referring to someone directly (ie try to write in the third person)?

tylerkelly13: fair. reworded.

PR #1175, opened 2019-08-23
TINY-3731: Added docs about new accessibility checker settings

lnewson:


plugins/a11ychecker.md
```diff
@@ -131,6 +131,52 @@ tinymce.init({
 });
 ```
 
+### `a11ychecker_html_version`
+
+This configuration option sets the HTML version to use when checking issues.
+
+For example, the "Complex tables should have summaries" rule in HTML4 the attribute is required because it's a valid attribute, however in HTML5 the attribute is deprecated, so the rule doesn't trigger.
```
wonderbugtiny: Maybe something like this?
>  For example, setting the version to HTML4 will trigger the rule "Complex tables should have summaries" as summary is a valid attribute and is required for tables, however in HTML5 the attribute is deprecated so, the rule wouldn't be triggered.

lnewson: Yeah, that's a lot better thanks!

plugins/a11ychecker.md
```diff
@@ -131,6 +131,52 @@ tinymce.init({
...
+    selector: "textarea",
+    plugins: "a11ychecker",
+    toolbar: "a11ycheck",
+    a11ychecker_html_version: 'html5'
+});
+```
+
+### `a11ychecker_level`
+
+This configuration option sets the WCAG level to use when checking issues.
```
tylerkelly13: `[WCAG level](https://www.w3.org/TR/WCAG20/#conformance)`

tylerkelly13: not critical, but helpful

PR #1181, opened 2019-08-28
Revert "Feature/pow 115"

tylerkelly13:
Reverts tinymce/tinymce-docs#1172
Delay in release of TinyMCE 5.0.14


tylerkelly13:
Not required

PR #1186, opened 2019-08-29
adding missing styles to the custom formats example and updating the …

tylerkelly13:
adding missing styles to the custom formats example and updating the documentation

### DIFF
--- a/_includes/configuration/content-style.md
+++ b/_includes/configuration/content-style.md
@@ -11,6 +11,18 @@ These styles are useful for previewing content as it appears outside the editor.
 ```js
 tinymce.init({
   selector: 'textarea',  // change this value according to your HTML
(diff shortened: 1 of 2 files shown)


_includes/configuration/content-style.md
```diff
@@ -11,6 +11,18 @@ These styles are useful for previewing content as it appears outside the editor.
 ```js
 tinymce.init({
-  content_style: "div {margin: 10px; border: 5px solid red; padding: 3px}"
+  content_style: 'div {margin: 10px; border: 5px solid red; padding: 3px; }'
```
wonderbugtiny: There should probably be a space in front of "margin" as well?

tylerkelly13: fixed

PR #1190, opened 2019-08-29
TINY-4055: Added 5.0.15 changelog

lnewson:


changelog.md
```diff
@@ -8,6 +8,13 @@ class: changelog
 
 {% capture changelog %}
 
+## Version 5.0.15 September 2, 2019
+* Added a dark content_css styles to go with the dark skin.
```
tylerkelly13: Added the dark content_css styles....?

changelog.md
```diff
@@ -8,6 +8,13 @@ class: changelog
...
+* Changed so the enabled state on toolbar buttons doesn't get the hover effect.
```
tylerkelly13: Changed the enabled state on toolbar buttons so they don't get the hover effect.  ?

changelog.md
lnewson: I think `a` is right here as it's not the one and only value, but maybe it should just be `style` instead of `styles` or even potentially `content_css skin`?

tylerkelly13: sounds good. had a feeling it was one of those :)

PR #1212, opened 2019-09-23
Ad-hoc docs release 2019-09-23

tylerkelly13:


_includes/integrations/react-tech-ref.md
```diff
@@ -0,0 +1,361 @@
...
+```xml
+<Editor
+  disabled={true}
+/>
+```
+
+#### `id`
+An id for the editor. Used for retrieving the editor instance using the `tinymce.get('ID')` method. Defaults to an automatically generated [UUID](https://tools.ietf.org/html/rfc4122).
+
+**Default value:** Automatically generated [UUID](https://tools.ietf.org/html/rfc4122).
```
EkimChau: in the example, we have a string 'uuid' as the placeholder, is that the default value for automatic UUID?

_includes/integrations/react-tech-ref.md
```diff
@@ -0,0 +1,361 @@
...
+
+**Default value:** `' '`
+
+**Type:** String
+
+##### Example: `initialValue`
+
+```xml
+<Editor
+  initialValue='Once upon a time...'
```
EkimChau: wrap in <p>tag</p> best practice

PR #1222, opened 2019-09-25
Improved the visuals of the mentions plugins example

lostkeys:
Please test in IE11. Both my parallels license and the virtual machine has expired 😞


lnewson:
Tested on IE 11 and looks fine, however we'll also need to update the full featured demo as it still uses the old styles.

lostkeys:
@tylerkelly13 can I leave the merge and pushing it out live to you?

PR #1235, opened 2019-10-02
editing the migration from 4 guide

tylerkelly13:


migration-from-4x.md
```diff
@@ -1,66 +1,76 @@
...
+| **Setting**                                                      | **TinyMCE 4**                                                                                 | **TinyMCE 5**                                                                                                                                                              |
+| ---------------------------------------------------------------- | --------------------------------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| [`height`]({{site.baseurl}}/configure/editor-appearance/#height) | Set the height of the editable area of the editor.<br/>Only accepted supported number values. | Sets the overall height of the editor, including the menubar, toolbars, and statusbar.<br/>Supports numbers and strings. Assumes strings are a valid CSS value for height. |
```
wonderbugtiny: is "accepted supported" the right phrase here?

migration-from-4x.md
```diff
@@ -1,66 +1,76 @@
...
-| Moderate | A [dialog]({{site.baseurl}}/ui-components/dialog/#dialoginstanceapi) created using `editor.windowManager.open` configuration objects | Convert TinyMCE 4.x config to TinyMCE 5.0 config. |
-| Major | Completely custom dialogs and extended use of the Modern UI framework | The new TinyMCE 5.0 components may not cover all API use cases. However, Tiny strives to create supported workarounds or if there are sufficient requests, add a new component to resolve the use case. |
+### Editor: Changes to API Methods
 
-> Note: For support related issues such as problems while migrating and a workaround, please contact [support](https://support.tiny.cloud/hc/en-us/requests/new). Alternatively, track developer preview issues on GitHub, [here](https://github.com/tinymce/tinymce/labels/dev%20preview).
+* All TinyMCE 4 API methods for creating UI components have been removed. New methods have been added for TinyMCE 5. For information on creating and customizing UI components, see: [User interface components]({{site.baseurl}}/ui-components/).
+* No core editor methods were removed (such as: `tinymce`, `editor`, `selection`, and `on()`).
```
lnewson: I realise this is how it was worded before, but this seems misleading/confusing, as only one of these are methods (`on()`) and it's more so an editor instance API. `tinymce` is a global and the entrypoint to the TinyMCE API, `editor` doesn't really exist and `selection` is an API available on editor instances.

migration-from-4x.md
```diff
@@ -146,37 +151,58 @@ The [Inlite]({{site.url}}/docs-4x/themes/inlite/) theme is no longer supported i
...
-
-The [Modern]({{site.url}}/docs-4x/themes/modern/)  theme is no longer supported in TinyMCE 5.0.  The modern theme's UI library [`tinymce.ui.*`]({{site.url}}/docs-4x/api/tinymce.ui/) has also been deleted. This change may impact integrations depending upon the [level of customization](#initialization).
-
-> Note: For support related issues such as problems while migrating and a workaround, please contact [support](https://support.tiny.cloud/hc/en-us/requests/new). Alternatively, track developer preview issues on GitHub, [here](https://github.com/tinymce/tinymce/labels/5.x).
+This will provide a similar but improved [distraction-free]({{site.baseurl}}/general-configuration-guide/use-tinymce-distraction-free/) experience in TinyMCE 5.
 
+### Modern theme removed
 
-### Events
+The [Modern]({{site.url}}/docs-4x/themes/modern/) theme is not supported in TinyMCE 5.  The Modern theme's UI library [`tinymce.ui.*`]({{site.url}}/docs-4x/api/tinymce.ui/) has also been removed. This change may impact integrations depending upon the level of customization.
```
lnewson: Another one that existed before that could be misleading? As we're saying it's removed, but then saying it's not supported which would imply it still exists?

migration-from-4x.md
```diff
@@ -146,37 +151,58 @@ The [Inlite]({{site.url}}/docs-4x/themes/inlite/) theme is no longer supported i
...
 | **Removed event** | **Description**|
 | ----------------- | -------------- |
 | BeforeRenderUi | Fired before the UI was rendered. |
 
 
-## Mobile support
+### Mobile theme
 
-In TinyMCE 4.x, [mobile support]({{site.baseurl}}/mobile/) was introduced bundled with a new theme and configuration settings. TinyMCE 5.0 makes this process seamless where mobile support comes out of the box without any additional configurations. The mobile theme will be automatically loaded if a user views the editor on a mobile phone. TinyMCE 5.0 mobile will be an exciting space to watch.
+The TinyMCE 4 [Mobile theme]({{site.baseurl}}/mobile/) is default behavior for the editor when it is loaded on a mobile device.
```
lnewson: `is the default`?

migration-from-4x.md
tylerkelly13: updated line 44 to: ```No core editor API methods were removed.```

migration-from-4x.md
tylerkelly13: Revised to ```Supported number values.```

migration-from-4x.md
tylerkelly13: Revised to ```has been removed from TinyMCE 5.```

migration-from-4x.md
tylerkelly13: Updated

migration-from-4x.md
```diff
@@ -413,23 +457,23 @@ Dialogs are still opened via the `editor.windowManager.open(config)` api, howeve
...
 
-##### Changed settings:
+##### Changed dialog settings:
 
 | **Old setting** | **New setting** | **Description** |
 | --------------- | --------------- | --------------- |
 | `onchange` | [`onChange`]({{site.baseurl}}/ui-components/dialog/#configurationoptions) | `onchange` now takes a callback function which is passed an API helper function and data. |
 
-> Note: `onchange`, a callback that was previously used to detect changes on each component, has been removed. The changes are now tracked by a single callback function with the change data. Refer to the [Interactive example using redial(config): void]({{site.baseurl}}/ui-components/dialog/#interactiveexampleusingredialconfigvoid) for more information.
+> Note: The `onchange` callback has been removed. Changes are now tracked using the redial callback function. For information on using redial, see: [Interactive example using redial]({{site.baseurl}}/ui-components/dialog/#interactiveexampleusingredial).
```
lnewson: This has changed the original meaning and isn't quite correct now (granted the original was confusing). Firstly, there is no such thing as a redial callback function. So the main difference between 4 and 5 was that in 4 there was a `onchange` function attached to each component, however in 5 there's only a single `onChange` callback at the root of the dialog spec, similar to `onAction`. The redial example linked contains an example to that, but it's more so about using the `redial` API to change the components/options in the dialog.

migration-from-4x.md
```diff
@@ -506,72 +575,63 @@ editor.windowManager.openUrl({
...
+* [Changed plugin settings](#changedpluginsettings)
+* [Changed plugins](#changedplugins)
+* [Spellchecker plugin](#spellcheckerplugin)
+* [Table plugin](#tableplugin)
 
-### Removed features
+### Removed plugin settings
 
-Plugins with dialogs no longer have `height` or `width` settings for their dialogs. The dialog component now uses CSS3 and a predefined `small`, `medium`, and `large` template to specify the dimensions.
+The `height` or `width` settings have been removed for plugin dialogs. The dialog component now uses CSS3 and a predefined `small`, `medium`, and `large` template to specify the dimensions.
```
lnewson: Should this be `and` instead of `or`?

migration-from-4x.md
```diff
@@ -506,72 +575,63 @@ editor.windowManager.openUrl({
...
 
-#### Changed plugin settings
+### Changed plugin settings
 
-| **TinyMCE 4.x** | **TinyMCE 5.0** |
-| [`autoresize_min_height`]({{site.url}}/docs-4x/plugins/autoresize/#autoresize_min_height) | [`min_height`]({{site.baseurl}}/configure/editor-appearance/#min_height) |
-| [`autoresize_max_height`]({{site.url}}/docs-4x/plugins/autoresize/#autoresize_max_height) | [`max_height`]({{site.baseurl}}/configure/editor-appearance/#max_height) |
-| [`textcolor_cols`]({{site.url}}/docs-4x/plugins/textcolor/#textcolor_cols) | [`color_cols`]({{site.baseurl}}/configure/content-appearance/#color_cols) |
-| [`textcolor_map`]({{site.url}}/docs-4x/plugins/textcolor/#textcolor_map) | [`color_map`]({{site.baseurl}}/configure/content-appearance/#color_map) |
+| **TinyMCE 4.x**                                                                           | **TinyMCE 5.0**                                                           |
```
lnewson: Should this be `TinyMCE 4` as I noticed `4.x` was changed to `4` in other places?

migration-from-4x.md
```diff
@@ -506,72 +575,63 @@ editor.windowManager.openUrl({
...
-For information on Context Toolbars, see the [docs]({{site.baseurl}}/ui-components/contexttoolbar).
-
 #### Custom context menus
 
-The [Context menu]({{site.url}}/docs-4x/plugins/contextmenu/) is no longer a plugin, it is part of the core and always enabled. Where TinyMCE 4.x only supported adding registered menu items, the new context menu also allows plugins to register "sections" of the context menu. These sections are dynamic and may show or hide depending on the cursor position when the context menu is opened. For example, the default context menu config is now `'link image editimage table spelling'` which are all plugin section references, not menu items. For information on using context menus, see: [UI components - Context menu]({{site.baseurl}}/ui-components/contextmenu/).
+The [Context menus]({{site.url}}/docs-4x/plugins/contextmenu/) are part of the core and enable by default in TinyMCE 5. TinyMCE 5 supports adding registered menu items and allows plugins to register "sections" of the context menu. These sections show or hide depending on the cursor position when the context menu is opened. The default context menu consists of plugin section references, not menu items.
```
lnewson: `s/enable/enabled`?

lnewson: Also should this be `default context menu configuration`? As it sounds like it's saying the when the context menu is opened it'll show sections instead of menu items atm.

migration-from-4x.md
lnewson: That's misleading as well though, as we did remove some of the `editor.windowManager` APIs?

lnewson: I'm wondering if we should just delete this line altogether?

migration-from-4x.md
tylerkelly13: Updated to:
```The `height` and `width` settings have been removed from plugin dialogs.```

migration-from-4x.md
tylerkelly13: Updated

migration-from-4x.md
tylerkelly13: It as a terrible sentence with little to no value that I tried to save.... and now it's gone.

PR #1239, opened 2019-10-04
Updating the Vue integration guide

tylerkelly13:
- Added Quick start guide
- Updated Tech reference content and formatting to improve consistence between the integration docs


_includes/integrations/vue-tech-ref.md
```diff
@@ -0,0 +1,367 @@
...
+
+#### `cloud-channel`
+
+**Default value:** `5`
+
+**Possible values:**  `5-stable`, `5-testing`, `5-dev`
+
+Changes the TinyMCE build used for the editor to one of the following cloud channels:
+
+- `5`, `5-stable` (**Default**): The current enterprise release of TinyMCE.
```
lnewson: Do we need default mentioned here again, given we have it above?

tylerkelly13: I'd like to leave it that way to reduce confusion

_includes/integrations/vue-tech-ref.md
```diff
@@ -0,0 +1,367 @@
...
+    * For bundle loader users (such as `webpack`, `rollup`, or `browserify`):
+
+        ```js
+        // es modules
+        import Editor from '@tinymce/tinymce-vue';
+        // commonjs require
+        // NOTE: default needed after require
+        var Editor = require('@tinymce/tinymce-vue').default;
+        ```
+    * To load `tinymce-vue` in a javascript file:
```
SimonFc: "html file"?

_includes/integrations/vue-tech-ref.md
```diff
@@ -0,0 +1,367 @@
...
+
+### Event binding
+
+Functions can be bound to editor events, such as:
+
+```html
+<editor @onSelectionChange="handlerFunction"></editor>
+```
+
+Where the `handlerFunction` is triggered with the event and returns an object containing two properties:
```
SimonFc: Same as in the tinymce-angular PR - maybe "and is called with two arguments" or something similar instead of "and returns an object containing two properties". Because it is `(e, editor) => void`, right?

_includes/integrations/vue-tech-ref.md
```diff
@@ -0,0 +1,367 @@
...
+- [Configuring the editor](#configuringtheeditor)
+
+    - [`api-key`](#api-key)
+    - [`cloud-channel`](#cloud-channel)
+    - [`disabled`](#disabled)
+    - [`id`](#id)
+    - [`init`](#init)
+    - [`initial-value`](#initial-value)
+    - [`inline`](#inline)
+    - [`modelEvents`](#modelevents)
```
SimonFc: `model-events`? Same in that section

_includes/integrations/vue-tech-ref.md
tylerkelly13: updated

_includes/integrations/vue-tech-ref.md
tylerkelly13: updated

_includes/integrations/vue-tech-ref.md
tylerkelly13: updated all instances of `modelEvents` to `model-events`

PR #1276, opened 2019-11-04
add sidebar issue to known issues for 5.1

tylerkelly13:


_includes/release-notes/51known-issues.md
```diff
@@ -28,4 +28,29 @@ To workaround this issue, users can either:
...
+#### Issue
+
+> **Note**: This issue affects the [Comments plugin]({{site.baseurl}}/plugins/comments/).
+
+When opening sidebars on small screens (such as mobile phones), the sidebar may occupy most or all of the text area of the editor.
+
+#### Workaround
+For the comments plugin, the comments sidebar can be collapsed using the `showcomments` toolbar button (![comment.svg]({{ site.baseurl }}/images/icons/comment.svg)).
+
+For other plugins with sidebars, no workaround exists. {{site.companyname}} recommends setting the `plugins` for `mobile` to exclude plugins with sidebars:
```
lnewson: Other plugins that use sidebars should be able to use the same workaround as comments, as when you register a sidebar TinyMCE will register a toolbar button with the same name. So the users would just need to make sure they include the toolbar button in their toolbar config.

tylerkelly13: updated

PR #1297, opened 2019-11-19
Advanced Tables documentation

tylerkelly13:
Did I miss anything?


_data/nav.yml
```diff
@@ -567,6 +568,7 @@
     - url: "configure"
     - url: "troubleshoot"
   - url: "advcode"
+  - url: "advanced_tables"
```
lnewson: This url is inconsistent with all other premium plugins (they use a hyphen instead of an underscore)

_includes/codepens/advtable-plugin/index.html
```diff
@@ -0,0 +1,160 @@
+<textarea id="advtable-plugin">
```
lnewson: The directory of this codepen and the id being used is a bit inconsistent, as all the others just use the name rather than adding the "-plugin" suffix. Not sure if you want to fix this, but just wanted to point it out.

enterprise/advanced_tables.md
```diff
@@ -0,0 +1,35 @@
+---
+layout: default
+title: Advanced Tables plugin
+title_nav: Advanced Tables
+description: Add sorting functionality to tables.
+keywords: sort tables advanced advtable
+---
+
+The initial release of Advanced Tables adds sorting functionality to tables in {{site.productname}}. Tables can be sorted by row or column values using:
```
lnewson: Should we have initial release mentioned here, as that feels like wording that should really only be in the release notes?

lnewson: Also it can now only be sorted by columns using the menus directly, so not sure if that could be confusing?

It's also possible to use the sort option in the table toolbar button.

enterprise/advanced_tables.md
```diff
@@ -0,0 +1,35 @@
...
+**For example:**
+
+| Sort rows based on the selected column                   | Advanced Sort Dialog                                 |
+| -------------------------------------------------------- | ---------------------------------------------------- |
+| ![Advanced Tables enhanced contextual menu for sorting rows based on the selected Column (Sort > Sort table by column ascending/descending).]({{ site.baseurl }}/images/advtable_context_menu_sort.png) | ![Advanced Tables sort dialog (Sort > Advanced Sort{{site.ellps}}).]({{ site.baseurl }}/images/advtable_dialog_sort.png) |
+
+### Try our Advanced Tables Demo
+{% include codepen.html id="advtable-plugin" %}
+
+### Getting started with {{site.productname}} Advanced Tables
```
lnewson: I thought we couldn't use the variables in headings as it breaks the anchor links?

enterprise/advanced_tables.md
```diff
@@ -0,0 +1,35 @@
...
+
+#### Creating an account
+
+To try out **Advanced Tables** and Tiny Cloud-delivered editor, the first step is to create a free [Tiny account](https://www.tiny.cloud/download/).
```
lnewson: This reads really awkwardly? The think it's the Tiny Cloud delivered editor part.

Also should this be using the cloud variable?

enterprise/advanced_tables.md
```diff
@@ -0,0 +1,35 @@
...
+
+The free user account is provisioned with a free 30-day trial of all Tiny [premium plugins](https://apps.tiny.cloud/product-category/tiny-cloud-extensions/), with no credit card information or commitment required.
+
+
+#### Buy {{site.productname}} Advanced Tables
```
lnewson: Again I didn't think we could use variables in headings?

plugins/advtable.md
```diff
@@ -0,0 +1,28 @@
+---
+layout: default
+title: Advanced Tables plugin
+title_nav: Advanced Tables
+description: Add sorting functionality to tables.
+keywords: advtable premium
```
lnewson: Maybe add the table keyword here?

plugins/advtable.md
```diff
@@ -0,0 +1,28 @@
...
+---
+
+The `advtable` plugin is a premium plugin that extends the core [`table` plugin]({{site.baseurl}}/plugins/table/) by adding sort options for rows and columns. Tables can be sorted by row or column values using:
+* The Sort options in the Table menu.
```
lnewson: There's also the table toolbar button that exposes sort.

_includes/codepens/advtable-plugin/index.html
lnewson: I wonder if we should move all plugins to a subdirectory, as that might be cleaner than using suffixes?

_data/nav.yml
tylerkelly13: updated

_includes/codepens/advtable-plugin/index.html
tylerkelly13: updated

tylerkelly13: we can move them next year

enterprise/advanced_tables.md
tylerkelly13: fixed

enterprise/advanced_tables.md
tylerkelly13: fixed

enterprise/advanced_tables.md
tylerkelly13: The whole thing is awkward really... but fixed?

enterprise/advanced_tables.md
tylerkelly13: Fixed

plugins/advtable.md
tylerkelly13: Added some keywords

plugins/advtable.md
tylerkelly13: added

PR #1301, opened 2019-11-25
Adding missing info to the template plugin docs

tylerkelly13:
needs verification, uncertain of `template_cdate_classes` example removing the `template_cdate_classes` line

### DIFF
--- a/plugins/template.md
+++ b/plugins/template.md
@@ -26,7 +26,7 @@ tinymce.init({
 
 | Name | Description |
 |------| ------------|
-| mceInsertTemplate | Insert template html |
+| `mceInsertTemplate` | Insert template html |
 
 ##### Example
 
@@ -36,7 +36,7 @@ tinymce.activeEditor.execCommand('mceInsertTemplate', false, '<p class="cdate">T
 
 There are two types of files the `template` plugin supports: *templates* and *snippets*. We will first review the plugin options before moving on to some examples.
 
-### Config Options
+### Configuration Options
 
 These settings affect the execution of the `template` plugin. Predefined templates for items such as created dates and modified dates can be set here.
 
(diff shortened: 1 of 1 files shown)


plugins/template.md
```diff
@@ -88,16 +102,26 @@ tinymce.init({
   plugins: "template",
   menubar: "insert",
   toolbar: "template",
-  template_cdate_classes: "cdate creationdate"
+  template_cdate_format: "%m/%d/%Y : %H:%M",
```
metricjs: I think this should go in the section below? This setting is for classes, not format

plugins/template.md
```diff
@@ -181,10 +209,46 @@ And that will be changed to:
...
+<p>Name: {$username}, StaffID: {$staffid}</p>
+```
+
+And the preview will look like:
+
+```html
+<p>Name: Jack Black, StaffID: 991234</p>
+```
+
 ### `template_selected_content_classes`
```
metricjs: http://fiddle.tinymce.com/gWgaab/1 <- see for example

plugins/template.md
tylerkelly13: fixed

plugins/template.md
```diff
@@ -173,35 +209,81 @@ This can then be used in a template or snippet that looks like this:
...
+
+And the preview will look like:
+
+```html
+<p>Name: Jack Black, StaffID: 991234</p>
+```
+
+### `template_selected_content_classes`
+
+When HTML elements in a template are assigned this content class, the content of the element will be replaced with selected content from the editor. This option accepts a list of content classes (separated by spaces).
```
metricjs: "content class" -> "class" might be a bit easier?

plugins/template.md
```diff
@@ -173,35 +209,81 @@ This can then be used in a template or snippet that looks like this:
...
+  templates: [
+    {
+      title: 'My Template',
+      description: 'This is my template.',
+      content: '<p>Hello, <span class="selcontent">this statement will be replaced.</span></p>'
+    }
+  ]
+});
+```
+If the word `world` is selected in the editor and _My Template_ is applied, `world` will be updated to:
```
metricjs: I was a little confused at first because there's no context for what the editor content is. But could also be because I made that fiddle, and this is similar enough that that's where my brain went first. Should probably get someone else to read it to check?

plugins/template.md
tylerkelly13: fixed

plugins/template.md
tylerkelly13: Fair enough.
I ran it by @thedrjim, who thinks it checks out.

PR #1304, opened 2019-11-27
fixes for the template plugin page

### LINKED ISSUE #492: Problem with page plugins/template.md
More example would be useful for custom template.

Missing also example of making snippet:
https://www.tinymce.com/docs/plugins/template/#makingsnippets

And here for custom variable there is not explained how to render them using {$varName}: https://www.tinymce.com/docs/plugins/template/#example

I just found on bottom under making snippet, but without that I was trying how to render variable "username" and "staffid" of the examples.

> *affected page: https://www.tinymce.com/docs/plugins/template/*

tylerkelly13:
Fixes #492

### DIFF
--- a/plugins/template.md
+++ b/plugins/template.md
@@ -323,15 +323,13 @@ This is the contents your backend page should return if you specify a URL in the
 ```json
 [
   {"title": "Some title 1", "description": "Some desc 1", "content": "My content"},
-  {"title": "Some title 2", "description": "Some desc 2", "url": "development.html"}
+  {"title": "Some title 2", "description": "Some desc 2", "url": "development.php"}
 ]
 ```
 
-### Making Templates
+### Making Template Files
 
-A template is a _file_ with a `div` containing the template data. All `html` outside the `div` will simply be presented to the user in the preview frame.
-
-Each template needs to be inside of a div with the `mceTmpl` class, like this example:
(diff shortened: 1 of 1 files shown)


plugins/template.md
```diff
@@ -353,8 +351,27 @@ Each template needs to be inside of a div with the `mceTmpl` class, like this ex
 
 ### Making Snippets
 
-Snippets are `html` code chunks that can be inserted. Replace variables will only be executed upon insert, without being wrapped in a template `div` element. So if you define `somevar1` in `template_replace_values` array it will be replaced on preview and insert.
+Snippets are `html` code chunks that can be inserted. Replace variables will only be executed upon insert, without being wrapped in a template `div` element. So if you define a variable in with [`template_replace_values`](#template_replace_values) it will be replaced on insert.
```
metricjs: "in with"?

plugins/template.md
```diff
@@ -329,9 +329,7 @@ This is the contents your backend page should return if you specify a URL in the
 
 ### Making Templates
```
metricjs: Should these headings become something like "Making Template Files" and "Making Inline Templates"? Idk what language we use elsewhere but these mean basically nothing to me

metricjs: And reference the content/url thing or whatever it is?

plugins/template.md
tylerkelly13: fixed

plugins/template.md
tylerkelly13: fixed

PR #1307, opened 2019-11-29
Improve the description of valid_styles

### LINKED ISSUE #646: Problem with page configure/content-filtering.md
The default is not specified. It's probably 'allow all', but it's unsure.

> *affected page: https://www.tinymce.com/docs/configure/content-filtering/*

tylerkelly13:
Fixes #646

### DIFF
--- a/_includes/configuration/valid-styles.md
+++ b/_includes/configuration/valid-styles.md
@@ -1,8 +1,8 @@
 ## valid_styles
 
-This option allows you to specify the available styles for each element. This option also forces the ordering of the styles. To use this option, specify the element name, then the allowed styles or `*` for all elements.
+To use this option, specify an object containing a mapping of element names to allowed styles. To specify the allowed styles for all elements, use `*` as the element name.
 
-By default, `valid_styles` is undefined, therefore all styles are valid. If a [`schema`](#schema) is set, `valid_styles` will override relevant parts of the `schema`.
+By default, all styles are allowed unless `valid_styles` or `invalid_styles` is configured.
 
 **Type:** `Object`
 
--- a/_includes/configuration/valid-styles.md
+++ b/_includes/configuration/valid-styles.md
@@ -1,6 +1,8 @@
 ## valid_styles
 
-This option enables you specify the available styles for each element. This means you can force ordering and only specific styles will be valid within style attribute values. Specify the element name, then it's styles or `*` for all elements.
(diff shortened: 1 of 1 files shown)


_includes/configuration/valid-styles.md
```diff
@@ -1,6 +1,8 @@
 ## valid_styles
 
-This option enables you specify the available styles for each element. This means you can force ordering and only specific styles will be valid within style attribute values. Specify the element name, then it's styles or `*` for all elements.
+This option allows you to specify the available styles for each element. This option also forces the ordering of the styles. To use this option, specify the element name, then the allowed styles or `*` for all elements.
+
+By default, `valid_styles` is undefined, therefore all styles are valid. If a [`schema`](#schema) is set, `valid_styles` will override relevant parts of the `schema`.
```
lnewson: The default at all times is to allow all styles, the only time that isn't the case is when `valid_styles` or `invalid_styles` is set. It really has nothing to do with the schema, as that only applies when manually instantiating `tinymce.dom.Styles`, but that doesn't apply to this setting.

lnewson: As such I think you can just say something like this instead maybe?

```By default all styles will be allowed, unless `valid_styles` or `invalid_styles` is configured.```

File: _includes/configuration/valid-styles.md [hunk 2]
```diff
@@ -1,6 +1,8 @@
 ## valid_styles
 
-This option enables you specify the available styles for each element. This means you can force ordering and only specific styles will be valid within style attribute values. Specify the element name, then it's styles or `*` for all elements.
+This option allows you to specify the available styles for each element. This option also forces the ordering of the styles. To use this option, specify the element name, then the allowed styles or `*` for all elements.
```
lnewson: This is a bit confusing to me, as is the asterisk referring to the element name or the styles list? I know it's the former but given it's directly after the mention of "styles" it sounds like the asterisk applies to the styles.

So may be something like this instead?
```
To use this option, specify an object containing a mapping of element names to allowed styles. To specify the allowed styles for all elements use `*` as the element name.
```

_includes/configuration/valid-styles.md
tylerkelly13: fixed

_includes/configuration/valid-styles.md
tylerkelly13: fixed

PR #1310, opened 2019-12-02
Add note about HTML3 to legacyoutput plugin

tylerkelly13:


plugins/legacyoutput.md
```diff
@@ -8,7 +8,9 @@ keywords: legacyoutput filter content cms email flash
 
 The `legacyoutput` plugin changes TinyMCE's output, producing legacy elements such as `font`, `b`, `i`, `u`, `strike`, and use `align` attributes.
 
-This plugin can be useful if you want to use TinyMCE in an HTML mail client or to render contents to Flash.
+> **Important**: This plugin changes the editor content to HTML3. Using this plugin will greatly reduce the functionality of the editor.
```
metricjs: I feel like "This plugin changes the editor content to HTML3" maybe doesn't give enough details, but I'm not sure what else it should say.

lnewson: Also does it actually change the content? I thought it only created new content as HTML 3, but I'm not certain on that.

lnewson: Looking more, it looks like it really only just makes it so that formatting (eg bold, italic, etc...) will be applied using HTML 3 markup and that's really all it does.

plugins/legacyoutput.md
```diff
@@ -8,7 +8,9 @@ keywords: legacyoutput filter content cms email flash
...
+
+This plugin can be useful if you want to use {{site.productname}} in an HTML mail client or to render contents to Flash.
```
lnewson: Also, I don't think this part is accurate anymore. Maybe it should say `legacy HTML mail clients`?

plugins/legacyoutput.md
tylerkelly13: Fixed?

plugins/legacyoutput.md
tylerkelly13: Fixed?

PR #1319, opened 2019-12-10
TinyMCE 5.1 documentation release

tylerkelly13:


_includes/codepens/basic-example/index.js
```diff
@@ -12,7 +12,7 @@ tinymce.init({
   ' alignright alignjustify | bullist numlist outdent indent |' +
   ' removeformat | help',
   content_css: [
-    '//fonts.googleapis.com/css?family=Lato:300,300i,400,400i',
-    '//www.tiny.cloud/css/codepen.min.css'
+  '//fonts.googleapis.com/css?family=Lato:300,300i,400,400i',
```
lnewson: The padding probably shouldn't have been removed here, as it's inside an array so it should be indented.

tylerkelly13: fixed by https://github.com/tinymce/tinymce-docs/pull/1319/commits/be9c330cb1666857f3d429453f8a8419212eb340

PR #1347, opened 2019-12-23
Updated hr.md

rmagon:
The plugin description says that it adds a toolbar button, so it feels like it adds the button by default. Instead, we need to add the 'hr' string to `toolbar`


tylerkelly13:
Thanks for fixing that up @rmagon!

PR #1349, opened 2020-01-08
TINY-3917: Added docs for the new placeholder setting

lnewson:


_includes/configuration/placeholder.md
```diff
@@ -0,0 +1,16 @@
+## placeholder
+
+This option adds placeholder content that will be shown when the editor is empty.
+
+Alternatively, the `placeholder` attribute on the original DOM element can be used instead, if the placeholder setting isn't specified.
```
EkimChau: nice!

spocke: Is the commas really needed here it's not an insertion?

tiny-james: I would rephrase it to:
`Alternatively, the \`placeholder\` attribute on the original DOM element will be used when the placeholder setting isn't specified.`

lnewson: I did consider that, but I wanted it to be phrased so that it's something the user should add if they want this, not that it's a fallback.

tylerkelly13: `If the placeholder setting isn't specified, the `placeholder` attribute on the original DOM element will be used.`
???

lnewson: That's exactly what I didn't want it to say though? It should be that you can add it not, that it'll fallback.

tylerkelly13: Collaborative effort from @tylerkelly13 @metricjs @lnewson 
`If the editor is initialized on a `textarea` element, the placeholder attribute can be used instead.`

tylerkelly13:
Do we know if this will be released with TinyMCE 5.1 or 5.2?

lnewson:
> Do we know if this will be released with TinyMCE 5.1 or 5.2? @lnewson ?

It'll be 5.2, so just updated the milestone.

tylerkelly13:
Thank @lnewson !

PR #1367, opened 2020-01-23
TINY-3328: Add docs for new link_default_protocol setting

ltrouton:


plugins/autolink.md
```diff
@@ -37,9 +37,31 @@ This option allows you to specify a default `target` value for links inserted by
...
+
+### `link_default_protocol`
+
+This option allows you to set a default protocol for links inserted by the autolink plugin.
+
+> **Note**: This setting also applies to the [link]({{ site.baseurl }}/plugins/link) plugin.
+
+**Type:** `String`
+
+**Default Value:** `'http'`
```
EkimChau: we should mention a notice: this default value will change to https as the default value in 6.0, somthing somthing best practices

PR #1383, opened 2020-02-03
Fixing first paragraph of the autosave plugin

tylerkelly13:


plugins/autosave.md
```diff
@@ -7,11 +7,10 @@ controls: toolbar button, menu item
 keywords: autosave_ask_before_unload autosave_interval autosave_prefix autosave_prefix autosave_restore_when_empty autosave_retention
 ---
 
-This plugin gives the user a warning if they made modifications to the content within an editor instance but didn't submit the changes. It also adds a menu item "Restore last draft" under the `File` menu. It also adds an optional button to the toolbar, which is useful if you don't intend to use menus.
+The autosave plugin gives the user a warning if they try to close the browser window or tab with unsaved changes in the editor. The autosave plugin adds:
```
metricjs: I think it's also if you try to navigate away, not just if you try to close things?

ltrouton: I think it also gives a warning on trying to refresh the page. Not sure if that needs to be put in though.

tylerkelly13: true. Tested on full featured demo. updating

tylerkelly13: updated, reworded the paragraph.

PR #1384, opened 2020-02-04
Remove references to MCFileManager

tylerkelly13:
And update the setting descriptions for relative_urls and remove_script_host

### DIFF
--- a/_includes/configuration/relative-urls.md
+++ b/_includes/configuration/relative-urls.md
@@ -1,6 +1,9 @@
 ## relative_urls
 
-If this option is set to true, all URLs returned from the MCFileManager will be relative from the specified [document_base_url](#document_base_url). If it's set to false all URLs will be converted to absolute URLs.
+For URLs with the same domain as the page containing the {{site.productname}} editor. If this option accepts the following values:
+
+- `true` --- All URLs created in {{site.productname}} will be converted to a link relative to the [`document_base_url`](#document_base_url).
+- `false` --- All URLs will be converted to absolute URLs.
 
 **Type:** `Boolean`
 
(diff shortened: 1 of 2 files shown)


_includes/configuration/relative-urls.md
```diff
@@ -1,6 +1,9 @@
 ## relative_urls
 
-If this option is set to true, all URLs returned from the MCFileManager will be relative from the specified [document_base_url](#document_base_url). If it's set to false all URLs will be converted to absolute URLs.
+For URLs with the same domain as the page containing the {{site.productname}} editor. If this option accepts the following values:
```
lnewson: The last sentence reads rather strangely. Im not sure if should be there?

metricjs: Should be "If set to:" I think?

_includes/configuration/remove-script-host.md
```diff
@@ -1,6 +1,8 @@
 ## remove_script_host
 
-If this option is enabled the protocol and host part of the URLs returned from the MCFileManager will be removed. This option is only used if the [relative_urls](#relative_urls) option is set to false.
+This option is used if the [`relative_urls`](#relative_urls) option is set to `false` and only applies to links with the same domain as the [`document_base_url`](#document_base_url).
+
+If this option is set to `false`, the `document_base_url` retained for relative links.
```
metricjs: This is worded weirdly and seems to be the opposite of the old explanation? I assume it's false by default, so shouldn't it say something like "If this option is set to true, the document_base_url is removed from relative links"?

tylerkelly13: default is true, so if not defined, the editor will remove the domain etc. the initial description was wrong based on my testing, can you verify from the source?

tylerkelly13: Clarified and fixed, thanks @metricjs

_includes/configuration/relative-urls.md
tylerkelly13: fixed

PR #1388, opened 2020-02-06
feat: new resize_img_proportional setting

tylerkelly13:


release-notes/release-notes52.md
```diff
@@ -53,9 +53,11 @@ For information on using the `TextSeeker` API, see: [{{site.productname}} APIs -
 
 ### Updated the colorinput component visual appearance --- updated color swatches in input fields
 
-### Missing resize_img_proportional setting
+### New proportional image resizing setting
```
lnewson: This isn't new and I believe it actually also controls table resizing. I'd double check with @jhaines though.

tylerkelly13: so have it just been fixed? revived?

lnewson: Fixed, as it broke back in I think 4.7.x.

tylerkelly13: release note removed

_includes/configuration/resize-img-proportional.md
```diff
@@ -0,0 +1,23 @@
+## resize_img_proportional
+
+When a user resizes an image in the editor, this option controls whether image is resized proportionally or freely. If set to:
+
+- `true` --- When users resize an image, the image will be resized proportionally (both dimensions of the image are resized by the same percentage of length).
```
EkimChau: we should mention, with the true setting, holding down shift while resizing will change the behaviour to free-form
and when set to false, holding shift while resizing will get proportional behaviours

jhaines: yeah this exactly

tylerkelly13: Updated with shift behaviour for both true and false options

PR #1389, opened 2020-02-06
feat: new page embed tiny_pageembed_inline_styles

tylerkelly13:


lnewson:
@tylerkelly13 I noticed the page embed page also documents the default styles. They got changed in 5.2, so we possibly need to update that as well. Here's the new styles:
```
.tiny-pageembed--21by9,
.tiny-pageembed--16by9,
.tiny-pageembed--4by3,
.tiny-pageembed--1by1 {
  display: block;
  overflow: hidden;
  padding: 0;
  position: relative;
  width: 100%;
}

.tiny-pageembed--21by9 {
  padding-top: 42.857143%;
}

.tiny-pageembed--16by9 {
  padding-top: 56.25%;
}

.tiny-pageembed--4by3 {
  padding-top: 75%;
}

.tiny-pageembed--1by1 {
  padding-top: 100%;
}

.tiny-pageembed--21by9 iframe,
.tiny-pageembed--16by9 iframe,
.tiny-pageembed--4by3 iframe,
.tiny-pageembed--1by1 iframe {
  border: 0;
  height: 100%;
  left: 0;
  position: absolute;
  top: 0;
  width: 100%;
}
```

plugins/pageembed.md
```diff
@@ -91,6 +91,28 @@ tinymce.init({
 });
 ```
 
+### Inline styles for page embeds
+
+The `tiny_pageembed_inline_styles` setting will inline all styles, instead of using CSS classes, when rendering the embed content. This is useful when the additional CSS classes can't be added to your site. Defaults to `false`.
```
jhaines: technically it keeps the classes as well, in this instance, unlike the other, so maybe "instead of" could be "while also keeping"?

- [ ] consider changing this?

tylerkelly13: done

PR #1392, opened 2020-02-11
Security Guide Updates

tylerkelly13:


advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
+  - [Cross-Site Scripting (XSS)](#cross-sitescriptingxss)
+  - [Injection](#injection)
+  - [Insecure Transmission and Storage of data](#insecuretransmissionandstorageofdata)
+  - [Broken Authentication and Session Management](#brokenauthenticationandsessionmanagement)
+  - [Using Components with Known Vulnerabilities](#usingcomponentswithknownvulnerabilities)
+  - [Further security information](#furthersecurityinformation)
+
+## Reporting TinyMCE security issues
+
+{{site.companyname}} values the work of security researchers in improving the security of technology products worldwide. We welcome researchers who wish to responsibly disclose vulnerabilities in our products or systems. Note that we do not offer any “bug bounty” program or any form of payment for disclosed vulnerabilities. If you would like to report a vulnerability, please email `infosec@tiny.cloud`.
```
lnewson: Maybe `[infosec@tiny.cloud](mailto:infosec@tiny.cloud)` so that it's a link instead of plain text?

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
+
+## What we do to maintain TinyMCEs' security
+
+To protect {{site.productname}}s' users, {{site.companyname}}:
```
lnewson: I don't really know why, but this just reads strangely... I think it needs something after "Tiny" like "To protect TinyMCEs’ users, Tiny performs the following:"

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
+
+- Patches Cross-Site Scripting (XSS) vulnerabilities,
+- Keeps {{site.productname}}s' dependencies up to date, and
+- Provides information how to configure a Content Security Policy that works with {{site.productname}}.
+
+### Scripts and XSS vulnerabilities
+
+{{site.productname}} filters out most Cross-Site Scripting (XSS) content such as scripts from the editor content. For additional security, {{site.companyname}} recommends passing the editor content through server-side filters such as [HTMLPurifier](http://htmlpurifier.org/).
```
lnewson: This isn't really for additional security, it's more so that any client-side only security can be circumvented, so they must do server-side filtering to ensure their published content remains safe.

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
+
+### Keeping dependencies up-to-date
+
+To protect our users, {{site.companyname}} ensures that TinyMCEs' dependencies are updated before the next version (major, minor, or patch) is released.
```
lnewson: We generally don't do dependency updates in a patch release, as that has a high risk of breaking things... we'd only do it if a security issue is raised in the dependency.

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
-| img-src 'self' *.tinymce.com *.tiny.cloud data: blob:; | Images within the editor are sometimes base64 encoded, blob URLs, or proxied through the {{site.cloudname}} service.
-| style-src 'self' 'unsafe-inline' *.tinymce.com *.tiny.cloud;        | Styles are used for inline formatting (such as underline, font colors, etc.) and the positioning of user interface elements.
-| font-src 'self' *.tinymce.com *.tiny.cloud;            | Fonts are used for icons in the UI and is loaded from external files.
+| Directives                                                   | Requirements                                                                                                                 |
+| ------------------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------- |
+| `script-src 'self' _.tinymce.com _.tiny.cloud;`                | Scripts are sometimes loaded as script element with an src attribute.                                                        |
```
lnewson: Why was the asterisks changed to underscores? as that's definitely not correct, they should be asterisks.

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
+
+The following security risks are not {{site.productname}} specific, but are the main security risks associated with websites or applications which allow user inputs. Protecting your services and users from these risks requires server-side handling. Note that attackers will likely bypass any editor and attack the server directly, rather than use the text editor as a vector.
+
+For information on mitigating these risks, see the provided links in each section.
+
+### Cross-Site Scripting (XSS)
+
+Cross-Site Scripting (XSS) involves using website or application inputs to inject malicious, client-side code. This code can then be used to attack your users.
+
+Although {{site.productname}} removes most XSS content, precautions should be taken to protect your users, such as enabling a Content Security Policy (CSP) as described in [Configuring Content Security Policy (CSP) for TinyMCE](#configuringcontentsecuritypolicycspfortinymce).
```
lnewson: I'd also add the HTML purify alternative you mentioned earlier, as not all users will be able to use a CSP. Even then it should be used in combination with a server-side santization library, as they could bypass the CSP and post directly to the server or since we require `unsafe-inline` for styles, an attacker could still trigger an XSS attack if TinyMCE was vulnerable.

So to sum up, we should be recommending a combination of things for preventing XSS, rather than the single CSP option (so basically the XSS/Sanitzation and Further information headings is all I'd keep).

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
 
 ```html
-<meta http-equiv="Content-Security-Policy" content="default-src 'none'; script-src 'self'; connect-src 'self' blob:; img-src 'self' data: blob:; style-src 'self' 'unsafe-inline'; font-src 'self';" />
+<meta
+  http-equiv="Content-Security-Policy"
+  content="default-src 'none'; script-src 'self'; connect-src 'self' blob:; img-src 'self' data: blob:; style-src 'self' 'unsafe-inline'; font-src 'self';"
+/>
 ```
+
+## Potential security risks and mitigation measures
```
lnewson: This whole section really feels out of place to be honest, as most of it is completely outside the scope of TinyMCE and really depends on how or where it's being integrated. As an example, if I was using TinyMCE in wordpress this would be rather confusing as they aren't implementing the things that would be affected themselves.

I'd suggest scrapping everything that doesn't impact TinyMCE directly and just include a heading that links to the OWASP 10 as things to be aware of.

advanced/security.md
tylerkelly13: I considered this, I decided not to link the email address to avoid automated spambots (somewhat). it is a security page after-all. I'm happy to link it if desired.

advanced/security.md
tylerkelly13: my bad, a vscode extension formatted the markdown. Will fix

advanced/security.md
tylerkelly13: As the first paragraph may indicate, I added this section as general information about potential risks related to user inputs, and to highlight that no editor will necessarily protect users/servers from attack.

lnewson: I still find it confusing having it like this though. Maybe we can break it out into a "Other security issues to be aware of outside TinyMCE" or similar?

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
-description_short: A statement on security.
-description: A statement on security.
-keywords: security xss scripting vulnerability hack hacker
+description_short: Security information for TinyMCE.
+description: Information on reporting security issues, what TinyMCE does to protect users, and what you can do to protect your users.
+keywords: security xss scripting vulnerability hack hacker csp mitigation protection protect
 ---
 
-## Q: Is TinyMCE protected against XSS vulnerabilities?
+> **Note**: The following is _general_ security advice that may be relevant to website or application using {{site.productname}}.
```
tinydylan: *a* website or application?

advanced/security.md
tinydylan: Yeah - sanitizing content server-side is essential. I'm not sure I would recommend a specific tool. I wouldn't say we filter out "most" XSS content - just say "we filter content such as scripts"

advanced/security.md
tinydylan: Grammar: apostrophe before s

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
-| style-src 'self' 'unsafe-inline' *.tinymce.com *.tiny.cloud;        | Styles are used for inline formatting (such as underline, font colors, etc.) and the positioning of user interface elements.
-| font-src 'self' *.tinymce.com *.tiny.cloud;            | Fonts are used for icons in the UI and is loaded from external files.
+| Directives                                                   | Requirements                                                                                                                 |
+| ------------------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------- |
+| `script-src 'self' _.tinymce.com _.tiny.cloud;`                | Scripts are sometimes loaded as script element with an src attribute.                                                        |
+| `connect-src 'self' _.tinymce.com _.tiny.cloud blob:;`         | XMLHttpRequest are required by some services such as spellchecking and PowerPaste.                                           |
```
tinydylan: *is* required?

advanced/security.md
tylerkelly13: updated to:
ensures that the TinyMCE dependencies are updated before the next version (major or minor) is released.

advanced/security.md
tylerkelly13: updated to:
> TinyMCE filters content such as scripts from the editor content, however, client-side applications can be by-passed by attackers. Tiny recommends processing received editor content through server-side filters.

advanced/security.md
tylerkelly13: fixed

advanced/security.md
tylerkelly13: changed to:
> precautions should be taken to protect your users, such as enabling a Content Security Policy (CSP) and server-side filtering.

advanced/security.md
tylerkelly13: renamed "General security risks for user input elements"

advanced/security.md
tylerkelly13: fixed

advanced/security.md
tylerkelly13: removed possessives from page

advanced/security.md
tylerkelly13: fixed

advanced/security.md
tylerkelly13: Updated to:
>To protect {{site.productname}} users, {{site.companyname}}:

advanced/security.md
```diff
@@ -2,46 +2,141 @@
...
+
+{{site.companyname}} values the work of security researchers in improving the security of technology products worldwide. We welcome researchers who wish to responsibly disclose vulnerabilities in our products or systems. Note that we do not offer any “bug bounty” program or any form of payment for disclosed vulnerabilities. If you would like to report a vulnerability, please email `infosec@tiny.cloud`.
+
+## What we do to maintain security for TinyMCE
+
+To protect {{site.productname}} users, {{site.companyname}}:
+
+- Patches Cross-Site Scripting (XSS) vulnerabilities,
+- Keeps the {{site.productname}} dependencies up to date, and
+- Provides information how to configure a Content Security Policy that works with {{site.productname}}.
```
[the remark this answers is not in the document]
metricjs: -> "Keeps TinyMCE dependencies..."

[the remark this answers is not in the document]
tylerkelly13: Thanks. Fixed

PR #1419, opened 2020-02-28
TINY-4688: TinyMCE lazy loading for integrations

SimonFc:


_includes/integrations/angular-quick-start.md
```diff
@@ -122,14 +114,40 @@ This procedure requires:
               { "glob": "**/*", "input": "node_modules/tinymce/plugins", "output": "/tinymce/plugins/" }
             ]
             ```
-        4. Update the editor configuration to include the `base_url` and `suffix` options.
+            
+            If you are using tinymce-angular 4.0.0 or later, add the {{site.productname}} core script as well
```
SimonFc: @jhaines The new feature documented here for tinymce-angular is yet to be released and with the following PR it would be part 4.0.0: https://github.com/tinymce/tinymce-angular/pull/140

lnewson: We released 3.5.0 instead since the original issue asked for it with the current version and we didn't want to force users to upgrade to get this. As such, we probably need to update the 2 references of `4.0.0` to `3.5.0`.

lnewson: Also this uses pronouns (you) so we probably should see if we can rework this a little.

lnewson: Also looking more, this really only applies to point 3a) below, so I don't think it makes sense to say to use it all the time. If we did want to use it all the time, then we might as well just include it in the above asset lists at all times.

lnewson: Actually, reading this more I have to wonder why we don't just say to do something like this instead:
```
"assets": [
  { "glob": "**/*", "input": "node_modules/tinymce", "output": "/tinymce/tinymce/" }
]
```
as that'll copy all the TinyMCE resources over.

_includes/integrations/angular-quick-start.md
```diff
@@ -122,14 +114,40 @@ This procedure requires:
...
+            ```js
+            import { EditorModule, TINYMCE_SCRIPT_SRC } from '@tinymce/tinymce-angular';
+            /* ... */
+            @NgModule({
+              /* ... */
+              imports: [
+                EditorModule
+              ],
+              providers: [
+                { provide: TINYMCE_SCRIPT_SRC, useValue: 'tinymce/tinymce.min.js' }
```
lnewson: This should probably be `/path/to/tinymce.min.js`, as that's what's used in other examples for `tinymceScriptSrc`. In saying that though, this likely be the correct value given the above step copies it across.

SimonFc:
@lnewson You're right about the comments you left, and thanks @tylerkelly13 for fixing the issues.

PR #1428, opened 2020-03-10
fix: missing linkchecker_preprocess setting

tylerkelly13:


plugins/linkchecker.md
```diff
@@ -60,14 +60,33 @@ a[data-mce-linkchecker-focus="true"] {
...
+**Type:** `Function`
+
+##### Example: `linkchecker_preprocess`
+
+```js
+tinymce.init({
+  selector: "textarea",  // change this value according to your HTML
+  plugins: "linkchecker",
+  linkchecker_preprocess: function (data) {
+    /* This example will double encode the url */
```
metricjs: Should we mention that data.url is already encoded, just to be super clear?

lnewson: It may not be though, as it's whatever link you give it generally?

tylerkelly13: changed to `/* This example will encode or double encode the url */`

PR #1437, opened 2020-03-23
5.2.1 changelog

tylerkelly13:


.api-version
```diff
@@ -1 +1 @@
-5.2.0
+5.2.1
```
lnewson: Should this have been bumped as I thought we normally did this when the enterprise goes out? Either way it's only your fixes to the API docs, so this should be okay.

changelog.md
```diff
@@ -10,6 +10,26 @@ class: changelog
 
 {% capture changelog %}
 
+## Version 5.2.1 March 25, 2020
+* Fixed the "is decorative" checkbox in the image dialog clearing after certain editor events.
```
lnewson: Just going to copy my comment from the TinyMCE PR here:
> This isn't really editor events, it's dialog events more so than anything.

.api-version
tylerkelly13: I'm ok with the API docs updating with community ATM

PR #1448, opened 2020-04-15
Remove comment about createObjectURL browser support

tinydylan:
This function is supported in all TinyMCE-supported browsers.


tinydylan:
See also: https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL

PR #1520, opened 2020-05-29
Doc fallback-to-get for linkcheck and consolidate

tylerkelly13:
Related Ticket: DOC-521

Description of Changes:
* Document new linkchecker server-side setting `fallback-to-get`
* Add release note and cross-ref premium plugin release notes to server-side changes
* Move linkchecker config setting to server-side component page and document undocumented `enabled` setting.
* Additional minor fixes.

DOD:

Review

### DIFF
--- a/enterprise/server/configure.md
+++ b/enterprise/server/configure.md
@@ -14,10 +14,7 @@ This configuration file will require you to enter *at least* the following  info
 
 - `allowed-origins` - the domains allowed to communicate with the server-side editor features. This is required by all server-side components.
 
-Some server-side components require additional configuration which can be found in their individual documentation:
-
-- [Enhanced Media Embed]({{ site.baseurl }}/enterprise/embed-media/mediaembed-server-config/)
-- [Link Checker]({{ site.baseurl }}/enterprise/check-links/#linkcheckersdkquicksetup)
+The Enhanced Media Embed server-side component require additional configuration, which can be found on the [Enhanced Media Embed page]({{ site.baseurl }}/enterprise/embed-media/mediaembed-server-config/).
 
 ### `allowed-origins` (required)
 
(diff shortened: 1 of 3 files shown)


enterprise/server/configure.md
```diff
@@ -216,6 +213,68 @@ ephox {
...
+    enabled = true
+  }
+}
+```
+
+#### `fallback-to-get` (optional)
+
+{{site.requires_jsscwar_230v}}
+
+The Link-checker normally relies on the `HEAD` response. If this setting is `true`, the link-checker will fall-back to `GET` responses. When `true`, the Link checker can correctly identify working URLs that return non-standard `HEAD` replies. This setting can lead to performance issues when enabled (`true`) and is set to `false` by default.
```
ephox-mogran, line 242: I think `this setting` loses the context of "this" too much. Probably restate the setting: 

> if `fallback-to-get` is true ....

ephox-mogran: >  This setting can lead to performance issues when enabled (`true`) and is set to `false` by default.

Would you normally give more justification than this? Or is this sufficient? Something like, 

Also, be careful about saying "performance issues" without specifying the client or the server. Performance issues on the server for something like link-checking will be noticeable to the integrator, performance issues on the client to the user. It won't affect the performance of the client (aside from how quickly they see underlines turn up)

ephox-mogran: > fallback (or fall-back, whatever we use consistently) to `GET` is also a bit misleading. It will do the `HEAD`, then when that fails for certain reasons, will do a `GET` as well to double-check. `fall-back` in the context you have it makes it sound a bit more like it just always uses GET from the beginning. Probably a problem with the name of the parameter :)

tylerkelly13: I've made a few changes to address these issues. 
Regarding:
> Would you normally give more justification than this? Or is this sufficient?

Sometimes. Bit of a balancing act between what users need to know and TMI. Keeping info minimal is important to save users time, and as the description grows, the likelihood of people reading it decreases.

ephox-mogran: That makes sense.

enterprise/server/configure.md
```diff
@@ -239,7 +239,7 @@ ephox {
 
 {{site.requires_jsscwar_230v}}
 
-The Link-checker normally relies on the `HEAD` response. If this setting is `true`, the link-checker will fall-back to `GET` responses. When `true`, the Link checker can correctly identify working URLs that return non-standard `HEAD` replies. This setting can lead to performance issues when enabled (`true`) and is set to `false` by default.
+The Link-checker normally relies on the `HEAD` response. If `fallback-to-get` is `true`, the link-checker may issue a `GET` request in addition to the `HEAD` request to verify a link. When `true`, the Link checker can correctly identify working URLs that return non-standard `HEAD` replies. Enabling the `fallback-to-get` setting can lead to server performance issues and is set to `false` by default.
```
ephox-mogran, line 242: Very close. In addition implies the possibility of parallelism. This is serial. If HEAD fails, then it tries GET.

tylerkelly13: > the link-checker may issue a `GET` request after the receiving a non-standard `HEAD` response to verify a link.

enterprise/server/configure.md
```diff
@@ -239,7 +239,7 @@ ephox {
 
 {{site.requires_jsscwar_230v}}
 
-The Link-checker normally relies on the `HEAD` response. If `fallback-to-get` is `true`, the link-checker may issue a `GET` request in addition to the `HEAD` request to verify a link. When `true`, the Link checker can correctly identify working URLs that return non-standard `HEAD` replies. Enabling the `fallback-to-get` setting can lead to server performance issues and is set to `false` by default.
+The Link-checker normally relies on the `HEAD` response. If `fallback-to-get` is `true`, the link-checker may issue a `GET` request after the receiving a non-standard `HEAD` response to verify a link. When `true`, the Link checker can correctly identify working URLs that return non-standard `HEAD` replies. Enabling the `fallback-to-get` setting can lead to server performance issues and is set to `false` by default.
```
ephox-mogran, line 242: after receiving or after the receiving of

tylerkelly13: oops. I changed that sentence a few times to get it right, left an artifact behind...
Fixed by: https://github.com/tinymce/tinymce-docs/pull/1520/commits/51661117cf90c568349e6411bd427b9d3dd71be8

PR #1521, opened 2020-05-29
new server-side http timeout settings

tylerkelly13:
Related Ticket: DOC-523

Description of Changes:
* Document new Alternative http timeout settings
* Add 5.3 release note

DOD:

Review

### DIFF
--- a/enterprise/server/configure.md
+++ b/enterprise/server/configure.md
@@ -197,6 +197,30 @@ ephox {
 }
 ````
 
+#### Alternative http timeout settings
+
+{{site.requires_jsscwar_230v}}
+
+When greater control over timeout settings is needed, the following three settings can be used instead of the `request-timeout-seconds` setting:
+
+- `connection-request-timeout-seconds`: The amount of time to wait for a connection from the connection pool.
+- `connect-timeout-seconds`: The amount of time to wait for a connection to be established.
+- `socket-timeout-seconds`: The amount of time to wait in between packets after a connection is established.
+
+If one or more of these three timeout settings are not configured, the services will the timeout length specified by `request-timeout-seconds`.
+
+For example:
+
+```
+ephox {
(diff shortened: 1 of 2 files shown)


enterprise/server/configure.md
```diff
@@ -197,6 +197,30 @@ ephox {
...
+
+{{site.requires_jsscwar_230v}}
+
+When greater control over timeout settings is needed, the following three settings can be used instead of the `request-timeout-seconds` setting:
+
+- `connection-request-timeout-seconds`: The amount of time to wait for a connection from the connection pool.
+- `connect-timeout-seconds`: The amount of time to wait for a connection to be established.
+- `socket-timeout-seconds`: The amount of time to wait in between packets after a connection is established.
+
+If one or more of these three timeout settings are not configured, the services will the timeout length specified by `request-timeout-seconds`.
```
ephox-mogran, line 210: There is a typo here.

ephox-mogran: A missing word. Probably make it more explicit as well.

This isn't quite right, actually

If `request-timeout-seconds` is present, it will be used, regardless of the presence of other settings.

If `request-timeout-seconds` is not present, then every setting (`connection-request-timeout-seconds`, `connect-timeout-seconds`, `socket-timeout-seconds`) must be present. They will have no defaulting logic.

>  Any  If one or more of these three timeout settings are not configured, the services will the timeout length specified by `request-timeout-seconds`.

tylerkelly13: Replaced with:

> If one of these settings are required, remove `request-timeout-seconds` and specify values for all three of these settings.

ephox-mogran: Looks good. Your example below using the same setting for all of them is a bit confusing to follow that piece of text, though.

PR #1522, opened 2020-05-29
regex allowed-origins docs and RN

tylerkelly13:
Related Ticket: DOC-522

Description of Changes:
* Added docs on using  regex for specifying allowed-origins for server-side components
* Added release note for regex allowed-origins support

DOD:

Review

### DIFF
--- a/enterprise/server/configure.md
+++ b/enterprise/server/configure.md
@@ -107,6 +107,24 @@ ephox {
 }
 ````
 
+#### Regular Expressions support
+
+{{site.requires_jsscwar_230v}}
+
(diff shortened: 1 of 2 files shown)


enterprise/server/configure.md
```diff
@@ -107,6 +107,24 @@ ephox {
 }
 ````
 
+#### Regular Expressions support
+
+
+Regular expressions can be used alongside [wildcards](#wildcardsupport) for specifying `allowed-origins.origins`. To use a regular expression, start and end the expression with the forward-slash `'/'` character.
```
ephox-mogran, line 114: It's worth double-checking, but all of this work was done before the decision to limit domains for a customer. I believe the limiting of domains is only cloud-only (and regex isn't supported for cloud), but it's worth checking that everyone knows this is coming out. Similar to the `same-origin` setting. Both settings make self-hosted origin limits kind of pointless.

PR #1536, opened 2020-06-03
Add documentation for the pastetext toolbar button for the Paste plugin

albinsunnanbo:
Related Ticket: 

Description of Changes:
Add documentation for the pastetext toolbar button for the Paste plugin

Pre-checks:

Review:

### DIFF
--- a/plugins/paste.md
+++ b/plugins/paste.md
@@ -24,7 +24,7 @@ tinymce.init({
   selector: "textarea",  // change this value according to your HTML
   plugins: "paste",
   menubar: "edit",
-  toolbar: "paste pastetext" // paste corresponds to the `Paste` menu button, pastetext corresponds to the `Paste as text` menu button
+  toolbar: "paste pastetext"
 });
 ```
 
--- a/plugins/paste.md
+++ b/plugins/paste.md
@@ -24,7 +24,7 @@ tinymce.init({
   selector: "textarea",  // change this value according to your HTML
   plugins: "paste",
   menubar: "edit",
-  toolbar: "paste"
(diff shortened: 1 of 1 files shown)


plugins/paste.md
```diff
@@ -24,7 +24,7 @@ tinymce.init({
   selector: "textarea",  // change this value according to your HTML
   plugins: "paste",
   menubar: "edit",
-  toolbar: "paste"
+  toolbar: "paste pastetext" // paste corresponds to the `Paste` menu button, pastetext corresponds to the `Paste as text` menu button
```
lnewson, line 27: The term `menu button` here is incorrect and actually means something else entirely. This should probably be `toolbar button` instead.

lnewson:
LGTM as well, with the exception of the incorrect term in the comment. Thanks for the contribution @albinsunnanbo!

plugins/paste.md
albinsunnanbo: I was referring to that the toolbar button is the same as the command in the menu, but the comment apparently could be interpreted in several ways.
Maybe I should just remove the comment altogether, the toolbar button names are pretty self explainable.
The main reason for the documentation update was to make it easy do discover the pastetext toolbar button as I had to dig a bit to find that there was such a button.

tylerkelly13: > Maybe I should just remove the comment altogether, the toolbar button names are pretty self explainable.

Might be the easiest option :)

albinsunnanbo: Done, 
It's could be one my smallest PR ever :-)

tylerkelly13:
Thanks again @albinsunnanbo. Update is now live on the documentation site: https://www.tiny.cloud/docs/plugins/paste/#example

PR #1677, opened 2020-09-21
add darkmode to selected demos

tylerkelly13:
Related Ticket: N/A

Description of Changes:
* Adding dark-mode to selected demos (based on device preferences).

Pre-checks:

Review:


_includes/live-demos/basic-example/index.js
```diff
@@ -11,5 +11,7 @@ tinymce.init({
   'bold italic backcolor | alignleft aligncenter ' +
   'alignright alignjustify | bullist numlist outdent indent | ' +
   'removeformat | help',
+  skin: (window.matchMedia('(prefers-color-scheme: dark)').matches ? 'oxide-dark' : ''),
```
lnewson, line 14: This seems as though it might be confusing to people getting started. Are we sure we want it on the basic demo?

Also IIRC an empty string is **not** the same as the variable not being set, so you should use `oxide` and `default` instead.

Lastly, instead of duplicating this line twice, I'd suggest moving it out. eg:

```
var useDarkMode = window.matchMedia('(prefers-color-scheme: dark)').matches;
...
skin: useDarkMode ? 'oxide-dark' : 'oxide',
content_css: useDarkMode ? 'dark' : 'default'
```

lnewson (changes requested):

_includes/live-demos/basic-example/index.js
tylerkelly13: :thinking: fair points. I'll leave this PR open for now, and fix it up later. I'll also fix the docs page while I'm here. 
FYI: https://www.tiny.cloud/docs/general-configuration-guide/customize-ui/#skins

tylerkelly13: Fixed in https://github.com/tinymce/tinymce-docs/pull/1677/commits/cf728e7029d1bf4deb2e1885c9b7fa4267aee3d7

lnewson (approved):

EkimChau (approved):

HAFRMO (approved):

PR #1788, opened 2020-12-03
Add summary to API docs data

metricjs:
Related Ticket: 

Description of Changes:
This allows us to add a paragraph at the top of an API docs page that contains more information than what fits/should go in the header. We can also use HTML in the summary, allowing for bold warnings, etc.

Pre-checks:
- [x] `_data/nav.yml` has been updated (if applicable) - NA
- [x] Files has been included where required (if applicable) - NA
- [x] Files removed have been deleted, not just excluded from the build (if applicable) - NA
- [x] (New product features only) Release Note added - NA

Review:


_includes/api-member.html
```diff
@@ -6,6 +6,14 @@ <h2><a class="anchorable" href="#extends">Extends</a></h2>
   {% endfor %}
 {% endif %}
 
+{% if include.data.summary.size > 0 %}
+  {% for summary in include.data.summary %}
+    {% unless include.data.desc contains summary %}
```
metricjs, line 11: Without this, for APIs that don't define a `summary` it will duplicate the `desc` (the info in the header) into the page, though it truncates it at the first fullstop. I attempted `!=` too, but that failed, I assume maybe because of the truncation.

_includes/api-member.html
```diff
@@ -6,6 +6,14 @@ <h2><a class="anchorable" href="#extends">Extends</a></h2>
   {% endfor %}
 {% endif %}
 
+{% if include.data.summary.size > 0 %}
```
lnewson, line 9: Why is this below the extends part? I would expect the summary to be first, with additional details below that.

metricjs: I didn't know what extends was, that's all. Will move it up.

[ ] Move

lnewson: It just basically says something like this:
```
Extends
Observable
```
I assume we don't use this, as it's horrible that isn't a list :facepalm:

metricjs: ...I've seen the tags used in a couple places, will investigate

metricjs: I don't think borrows/Extends is working anyway.

ltrouton (approved):

JamesToohey (approved):

EkimChau (approved):

lnewson (approved):

_includes/api-member.html
tylerkelly13: truncation should only occur on the "index" pages (the ones with the "tiles")

tylerkelly13 (approved):

PR #1820, opened 2021-01-04
add supported versions to angular page

tylerkelly13:
Related Ticket: N/A

Description of Changes:
* add supported versions to angular page

Pre-checks:

Review:

### DIFF
--- a/_includes/integrations/angular-tech-ref.md
+++ b/_includes/integrations/angular-tech-ref.md
@@ -2,6 +2,7 @@
 
 **Covered in this section:**
 
+- [Supported Angular versions](#supportangularversion)
 - [Installing the TinyMCE Angular integration using NPM](#installingthetinymceangularintegrationusingnpm)
 - [Using the TinyMCE Angular integration](#usingthetinymceangularintegration)
 - [Configuring the editor](#configuringtheeditor)
@@ -30,6 +31,16 @@
   - [`allowedEvents`](#allowedevents)
   - [`ignoreEvents`](#ignoreevents)
 
+### Supported Angular versions
+
+The following table shows the supported versions of Angular and the required version of the `tinymce-angular` integration package.
+
+| Angular Version | `tinymce-angular` version |
+|:---:|:---:|
+| 9+ | 4.x |
+| 6 to 8 | 3.x |
+| 5 or older | Not Supported |
+
 ### Installing the TinyMCE Angular integration using NPM
 
 To install the `tinymce-angular` package and save it to your `package.json`.
@@ -37,8 +48,6 @@ To install the `tinymce-angular` package and save it to your `package.json`.
 ```sh
 $ npm install --save @tinymce/tinymce-angular
 ```
-
-This package is for Angular 5+. For Angular 4, use `@tinymce/tinymce-angular-legacy`
 
 ### Using the TinyMCE Angular integration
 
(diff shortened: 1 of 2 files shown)


_includes/integrations/angular-tech-ref.md
```diff
@@ -2,6 +2,7 @@
 
 **Covered in this section:**
 
+- [Supported Angular versions](#supportangularversion)
```
JamesToohey, line 5:
```suggestion
- [Supported Angular versions](#supportedangularversions)
```

JamesToohey: I assume this needs to match the heading to work?

JamesToohey (approved):

techtangents (approved):

_includes/integrations/angular-tech-ref.md
```diff
@@ -30,6 +31,16 @@
...
   - [`ignoreEvents`](#ignoreevents)
 
+### Supported Angular versions
+
+The following table shows the supported versions of Angular and the required version of the `tinymce-angular` integration package.
+
+| Angular Version | `tinymce-angular` version |
+|:---:|:---:|
+| 9+ | 4.x |
+| 6 to 8 | 3.x |
```
lnewson, line 41: This seems incorrect... I'm 99% sure this should be `5 to 8` as 3.x always supported that range previously and then we have the `@tinymce/tinymce-angular-legacy` package for 4.x.

As a side note, should the `@tinymce/tinymce-angular-legacy` package also be mentioned below? as I find it rather confusing to say not supported and then have the page this be used in say `Angular 4+ integration`

lnewson (changes requested):

_includes/integrations/angular-tech-ref.md
lnewson: Hmm... I missed you changed the title to 6+. However there are still other places that refer to using the legacy package for Angular 4.

_includes/integrations/angular-tech-ref.md
tylerkelly13: yeah, they do :) Thanks

_includes/integrations/angular-tech-ref.md
tylerkelly13: Just checked the source, we only mentioned the legacy package once (the line I removed)

lnewson: That still doesn't address the incorrect versions listed there... also more needs to happen if we're going to drop the "legacy" package (not from the docs side, but for example the package needs to be marked deprecated, etc...).

lnewson (changes requested):
The incorrect `6+` still hasn't been fixed, as that needs to be `5+` at the very least.

_includes/integrations/angular-tech-ref.md
tylerkelly13: Fixed by https://github.com/tinymce/tinymce-docs/pull/1820/commits/2097dc66603756fc01c1945969537ffaa7233aef

tylerkelly13:
> The incorrect `6+` still hasn't been fixed, as that needs to be `5+` at the very least.

after internal discussion, we have clarified that Angular 5+ is the minimum supported version. This issue has been corrected by https://github.com/tinymce/tinymce-docs/pull/1820/commits/2097dc66603756fc01c1945969537ffaa7233aef

lnewson (approved):

PR #1828, opened 2021-01-09
DOC-743: Enhancement note for improved copy pasting of tables

JamesToohey:
Related Ticket: DOC-743 & TINY-6664

Description of Changes:
Added release note for enhancement to table plugin when copy/pasting tables

Pre-checks:
- [x] ~~`_data/nav.yml` has been updated (if applicable)~~
- [x] ~~Files has been included where required (if applicable)~~
- [x] ~~Files removed have been deleted, not just excluded from the build (if applicable)~~

Review:

### DIFF
--- a/release-notes/release-notes57.md
+++ b/release-notes/release-notes57.md
@@ -37,7 +37,9 @@ The following new features were added for the {{site.productname}} 5.7 release.
 
 The following enhancements were made for the {{site.productname}} 5.7 release.
 
-### Enhancement name
+### Table widths are now retained when copying and pasting tables
+
+Tables copied to the clipboard when using the `table` plugin will now retain their width and the width of each column when pasted.
 
 ### Additional enhancements
 
--- a/release-notes/release-notes57.md
+++ b/release-notes/release-notes57.md
@@ -39,7 +39,7 @@ The following enhancements were made for the {{site.productname}} 5.7 release.
 
 ### Table widths are now retained when copying and pasting tables
 
-Tables copied to the clipboard when using the `table` plugin will now retain their width and the width of each column when pasted.
+Tables copied to the clipboard when using the `table` plugin will now correctly retain the width of each column. Additionally, the table width will be correctly adjusted to account for the copied column widths.
 
 ### Additional enhancements
 


release-notes/release-notes57.md
```diff
@@ -37,7 +37,9 @@ The following new features were added for the {{site.productname}} 5.7 release.
 
 The following enhancements were made for the {{site.productname}} 5.7 release.
 
-### Enhancement name
+### Table widths are now retained when copying and pasting tables
+
+Tables copied to the clipboard when using the `table` plugin will now retain their width and the width of each column when pasted.
```
lnewson, line 42: This doesn't quite sound right to me. The table width won't be retained, instead it'll be adjusted to match the size of the copied columns. So maybe something like this instead?

```suggestion
Tables copied to the clipboard, when using the `table` plugin, will now correctly retain the width of each column and adjust the table width to match the column widths when pasted.
```

JamesToohey: This was a difficult one. I chose the wording based on the assumption that the subset of the rows/columns that you've copied are (from an ultra generalised perspective) a new table when you copy them.

I agree it needs work but I don't think this works either, particularly: "adjust the table width to match the column widths when pasted." because technically the table width is the sum of the column widths.

Needs more work, for sure, I'm keen for @tylerkelly13's advice.

metricjs:
```suggestion
Tables copied to the clipboard when using the `table` plugin will now correctly retain the width of each column. When pasted into another table, its width will be adjusted to account for the column widths.
```
or something? not great either but I thought that splitting the copying part and pasting part into two sentences might make things a little clearer?

lnewson: That's not quite correct either, as the table width is adjusted on copy.

JamesToohey:
```suggestion
Tables copied to the clipboard when using the `table` plugin will now correctly retain their width and the widths of each of the copied columns.
```

lnewson: That has the same issue that I mentioned in the original comment, as it sounds like the table will remain the same size.

lnewson:
```suggestion
Tables copied to the clipboard when using the `table` plugin will now correctly retain the width of each column. Additionally, the table width will be adjusted to account for the copied column widths.
```
Maybe?

JamesToohey: I'm happy with that ^

tylerkelly13 (approved):

lnewson (approved):

EkimChau (approved):

PR #1928, opened 2021-04-13
fixing tiny.cloud urls that have changed

tylerkelly13:
Related Ticket: N/A

Description of Changes:
* Fixing some links that moved during product site refresh

Pre-checks:

Review:

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -37,7 +37,7 @@ enterpriseplan: Tiny Custom Plans
 supporturl: https://support.tiny.cloud
 plugindirectory: https://www.tiny.cloud/tinymce/features/
 gettiny: https://www.tiny.cloud/get-tiny
-accountpageurl: https://www.tiny.cloud/my-account/
+accountpageurl: https://www.tiny.cloud/my-account
 accountsignup: https://www.tiny.cloud/auth/signup
 productpages: https://apps.tiny.cloud/products
 pricingpage: https://www.tiny.cloud/pricing
@@ -106,4 +106,3 @@ plugins:
   - jekyll-last-modified-at
   - jekyll-redirect-from
   - jekyll-sitemap
-
--- a/_config.yml
+++ b/_config.yml
@@ -35,9 +35,9 @@ tierthreeplan: Tiny Cloud Professional Plan
 enterpriseplan: Tiny Custom Plans
 ## common tiny.cloud URLs
 supporturl: https://support.tiny.cloud
-plugindirectory: https://www.tiny.cloud/apps/
+plugindirectory: https://www.tiny.cloud/tinymce/features/
 gettiny: https://www.tiny.cloud/get-tiny
-accountpageurl: https://apps.tiny.cloud/my-account
+accountpageurl: https://www.tiny.cloud/my-account/
 accountsignup: https://www.tiny.cloud/auth/signup
 productpages: https://apps.tiny.cloud/products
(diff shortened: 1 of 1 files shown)


_config.yml
```diff
@@ -35,9 +35,9 @@ tierthreeplan: Tiny Cloud Professional Plan
 enterpriseplan: Tiny Custom Plans
 ## common tiny.cloud URLs
 supporturl: https://support.tiny.cloud
-plugindirectory: https://www.tiny.cloud/apps/
+plugindirectory: https://www.tiny.cloud/tinymce/features/
 gettiny: https://www.tiny.cloud/get-tiny
-accountpageurl: https://apps.tiny.cloud/my-account
+accountpageurl: https://www.tiny.cloud/my-account/
```
tylerkelly13, line 40:
```suggestion
accountpageurl: https://www.tiny.cloud/my-account
```

tylerkelly13: just in case the `/` was missing for a reason.

ltrouton (approved):

ashfordneil (approved):

_config.yml
```diff
@@ -35,9 +35,9 @@ tierthreeplan: Tiny Cloud Professional Plan
 enterpriseplan: Tiny Custom Plans
 ## common tiny.cloud URLs
 supporturl: https://support.tiny.cloud
-plugindirectory: https://www.tiny.cloud/apps/
+plugindirectory: https://www.tiny.cloud/tinymce/features/
 gettiny: https://www.tiny.cloud/get-tiny
-accountpageurl: https://apps.tiny.cloud/my-account
+accountpageurl: https://www.tiny.cloud/my-account
```
lnewson, line 40: This doesn't look correct, as it's immediately redirected to `https://www.tiny.cloud/my-account/` so shouldn't it be that instead?

lnewson (approved):

_config.yml
tylerkelly13: I'll check where we use this attribute to make sure it wont break anything. typically we use these URLs as `{{site.attribute}}/page/` or `{{site.attribute}}` and let it redirect to the slash version

lnewson: ahh, right I forgot about that sorry :cry:

tylerkelly13: no worries. I'll leave as is. (it's easier to read in text as-is)

_config.yml
```diff
@@ -35,9 +35,9 @@ tierthreeplan: Tiny Cloud Professional Plan
...
 accountsignup: https://www.tiny.cloud/auth/signup
 productpages: https://apps.tiny.cloud/products
```
EkimChau, line 42: would this one day need to be https://www.tiny.cloud/products/

_config.yml
EkimChau, line 40: we should add trailing slashes/ for all the links, it should save on 301 redirects

EkimChau (approved):

_config.yml
tylerkelly13: hopefully. But for now that is correct.

_config.yml
tylerkelly13: in the current system, baseurl and url are set by the site, such as:
```html
<body role="document" data-shared-baseurl="" data-baseurl="/docs">
```
so if we do this, some will have the trailing slash, others wont; which could get confusing. Hopefully we will eventually fix this.

PR #1967, opened 2021-05-14
TINY-7431: Added additional spellchecker release notes and fix issues with spellchecker docs

lnewson:
Related Ticket: TINY-7431

Description of Changes:
* Added additional bug fixes to release notes
* Fixed various issues in the spellchecker docs (incorrect/inconsistent languages and default values)

Pre-checks:

Review:

### DIFF
--- a/_includes/configuration/languages.md
+++ b/_includes/configuration/languages.md
@@ -4,10 +4,10 @@ The following languages are supported:
 
 | **Language**                                 | **Code**            |
 |----------------------------------------------|---------------------|
-| English (US)                                 | en, en-US           |
-| English (UK)                                 | en-UK, en-GB, en-BR |
-| English (US) - with additional medical terms | en-US-medical       |
-| English (UK) - with additional medical terms | en-GB-medical       |
+| English (US)                                 | en, en_US           |
+| English (UK)                                 | en_UK, en_GB, en_BR |
+| English (US) - with additional medical terms | en_US-medical       |
+| English (UK) - with additional medical terms | en_GB-medical       |
 | Danish                                       | da                  |
 | Dutch                                        | nl                  |
 | Finnish                                      | fi                  |
@@ -16,6 +16,6 @@ The following languages are supported:
 | Italian                                      | it                  |
 | Norwegian                                    | nb                  |
 | Portuguese (Brazil)                          | pt                  |
-| Portuguese (Europe)                          | pt-PT               |
+| Portuguese (Europe)                          | pt_PT               |
 | Spanish                                      | es                  |
 | Swedish                                      | sv                  |
--- a/_includes/configuration/languages.md
+++ b/_includes/configuration/languages.md
@@ -5,7 +5,7 @@ The following languages are supported:
 | **Language**                                 | **Code**            |
 |----------------------------------------------|---------------------|
 | English (US)                                 | en, en-US           |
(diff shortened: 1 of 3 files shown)


plugins/premium/tinymcespellchecker.md
```diff
@@ -141,7 +141,7 @@ This option specifies the spellchecker languages that are available to the user,
 tinymce.init({
   selector: 'textarea',
   plugins: 'tinymcespellchecker',
-  spellchecker_languages: 'US English=en_us,UK English=en_gb,Danish=da,Dutch=nl,Finnish=fi,French=fr,German=de,Italian=it,Norwegian=nb,Brazilian Portuguese=pt,Iberian Portuguese=pt_PT,Spanish=es,Swedish=sv'
+  spellchecker_languages: 'US English=en-US,UK English=en-GB,Danish=da,Dutch=nl,Finnish=fi,French=fr,German=de,Italian=it,Norwegian=nb,Brazilian Portuguese=pt,Iberian Portuguese=pt-PT,Spanish=es,Swedish=sv'
```
HAFRMO, line 144: Shouldn't the `US English` and `UK English` match the changes you did above?

lnewson: No, the demo is supposed to show setting a different value to the defaults.

tylerkelly13 (approved):

_includes/configuration/languages.md
```diff
@@ -4,10 +4,10 @@ The following languages are supported:
 
 | **Language**                                 | **Code**            |
 |----------------------------------------------|---------------------|
-| English (US)                                 | en, en-US           |
-| English (UK)                                 | en_uk, en_gb, en_br |
-| English (US) - with additional medical terms | en-US-medical       |
-| English (UK) - with additional medical terms | en-GB-medical       |
+| English (US)                                 | en, en_US           |
+| English (UK)                                 | en_GB, en_UK, en_BR |
+| English (US) - with additional medical terms | en_US-medical       |
```
ashfordneil, line 9: Is it worth a "requires 5.8" warning here? I know this doesn't fit within what we normally use that warning for, but I feel like we might want to show a "this didn't really work for a long time" caveat in the docs for this setting, and that's how we normally show those warnings elsewhere

ashfordneil (approved):

_includes/configuration/languages.md
lnewson: That would imply it's a feature added then, which wouldn't fit it's normal use case so I think we'd need a different warning if we wanted something here.

lnewson: It's also a tricky one, because up until a certain version of SpellChecker Pro it did work, so it more needs to be a certain version range which would be weird.

We also don't normally document things that were broken between certain version ranges like this, so my vote would be to not add anything here.

tylerkelly13: technically, isn't it the spellchecker version, not tinymce version? the spellchecker version we are releasing should be (somewhat) backward compatible through 5.X?

lnewson: That is correct also :+1:

tylerkelly13: verified internally. These are primarily docs fixes. Adding version information is not appropriate as the dictionaries were not added in this version, a bug prevented them from working as expected in previous versions.

EkimChau (approved):

PR #1987, opened 2021-05-27
DOC-831: removed passphrase since it makes it hard to be used with some of the JWT libraries

spocke:
Related Ticket: 

Description of Changes:
* Removing the passphrase part since that is not really needed to get things going.
* Tested these steps on linux, mac and windows.

Pre-checks:

Review:

### DIFF
--- a/_data/nav.yml
+++ b/_data/nav.yml
@@ -682,7 +682,7 @@
       - url: "#JWT endpoint setup procedure"
       - url: "#PHP token provider endpoint example"
       - url: "#Node token provider endpoint example"
-      - url: "#Creating a private/public key pair for RTC"
+      - url: "#Creating a private/public key pair for Real-Time Collaboration (RTC)"
     - url: "encryption"
       pages:
       - url: "#Encryption keys are mandatory"
--- a/_includes/auth/gen-rsa-key-pairs.md
+++ b/_includes/auth/gen-rsa-key-pairs.md
@@ -65,13 +65,15 @@ sudo zypper install openssl
 
 To create a private/public key pair:
 
-On a command line, run:
+1. On a command line, run:
 
-```sh
-ssh-keygen -m PEM -t rsa -b 2048 -N "" -f <MY_KEY_PAIR_NAME>
-```
+    ```sh
+    ssh-keygen -m PEM -t rsa -b 2048 -f <MY_KEY_PAIR_NAME>
+    ```
 
-Where _`<MY_KEY_PAIR_NAME>`_ should be replaced with a name for the key pair.
+    Where _`<MY_KEY_PAIR_NAME>`_ should be replaced with a name for the key pair.
+
+2. Enter a passphrase for accessing the key.
 
 Two files will be created in the current directory:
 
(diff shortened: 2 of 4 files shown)


_includes/auth/gen-rsa-key-pairs.md
```diff
@@ -65,15 +65,13 @@ sudo zypper install openssl
 
 To create a private/public key pair:
 
-1. On a command line, run:
+On a command line, run:
```
metricjs, line 68:
```suggestion
On a command line run the following, where _`<MY_KEY_PAIR_NAME>`_ should be replaced with a name for the key pair:
```
instead of splitting the sentence?

JamesToohey (approved):

_includes/auth/gen-rsa-key-pairs.md
lnewson: I wonder if we need even more than this? As having the `To create` thing above makes it look like we're going to mention a list but then we don't :shrug:

lnewson (approved):

_data/nav.yml
```diff
@@ -682,7 +682,7 @@
       - url: "#JWT endpoint setup procedure"
       - url: "#PHP token provider endpoint example"
       - url: "#Node token provider endpoint example"
-      - url: "#Creating a private/public key pair for RTC"
+      - url: "#Creating a private/public key pair for Real-Time Collaboration (RTC)"
```
tylerkelly13, line 685: Sorry. I just noticed that I broke this the other week (I think).

_includes/auth/gen-rsa-key-pairs.md
tylerkelly13: Whatever the decision was, it was the right one, 'cos it looks good

tylerkelly13:
> Removing the passphrase part since that is not really needed to get things going

true, just wondering if we should add a note to say that they add a passphrase where possible for production deployments?

_includes/auth/gen-rsa-key-pairs.md
```diff
@@ -63,17 +63,13 @@ sudo zypper install openssl
...
 #### Create a private/public key pair on Linux
 
-To create a private/public key pair:
-
-1. On a command line, run:
-
-    ```sh
-    ssh-keygen -m PEM -t rsa -b 2048 -f <MY_KEY_PAIR_NAME>
-    ```
+{% include auth/recommend-passphrase.md %}
```
tylerkelly13, line 66: Hi @spocke, I've added a notice to the relevant sections, _partly_ for legal reasons. We need to be clear that this is not ideal for security, and do not want to be encouraging poor practices.

spocke:
> > Removing the passphrase part since that is not really needed to get things going
> 
> true, just wondering if we should add a note to say that they add a passphrase where possible for production deployments?

I think it will just complicate the setup. Our generated keys from the account page doesn't produce keys with a passphrase and since some of the JWT libraries doesn't even support that it will be a support issue if people start asking for that or how to set that up.

_includes/auth/gen-rsa-key-pairs.md
spocke: Is it though. We generate a unique private/public key pair by cryptographically random generation we tell them to keep the private key safe. Adding a passphrase to my mind adds no extra security since you would have to pass that passphrase in anyway and have it stored possibly together with the key. You might be able to have a setup where the key is stored in one place then the passphrase in another but they would still need to come together to produce a signed JWT key and that would be the attack surface where you would have both things anyway. To me passphrases is more if you use the key for login purposes where you want a user to enter that passphrase in some secure context. @tjdett might know more about this.

tylerkelly13: I've had a think on this, and i may have a better solution that better addresses both our concerns. I'll revert my commit and make fresh edits tomorrow.

tylerkelly13: Ok,so I've revised the approach. Rather than tell them to add or not add a passphrase, we will indicate that the passphrase is optional and follow it with a note indicating why.

tjdett (approved):

EkimChau (approved):

PR #2027, opened 2021-07-12
DOC-844: Adding ignore-port details

ephox-mogran:
Related Ticket: DOC-844

Description of Changes:
* adding information about ignore-port

Pre-checks:
- [X] Branch prefixed with `feature/` or `hotfix/`
- [X] `_data/nav.yml` has been updated (if applicable)

Review:


enterprise/server/troubleshoot.md
```diff
@@ -28,7 +28,7 @@ If the editor is reporting that the service cannot be found and tracing the netw
 
 ### Chrome specific tips
 
-If the server is not running on a standard HTTP or HTTPS port (80 or 443) then Chrome will include the port in the Origin header that is sent to the server. Other browsers do not do this. This is why when specifying the `allowed-origins` config, you should use both the hostname by itself and the hostname and port in the configuration. See [Additional Information Around Entering Origins]({{ site.baseurl }}/enterprise/server/#additionalinformationaroundenteringorigins) for more details.
+If the server is not running on a standard HTTP or HTTPS port (80 or 443) then Chrome will include the port in the Origin header that is sent to the server. Other browsers do not do this. If you have set `ignore-port` to be `false`, then you will need to include the port in the `allowed-origins`. See [Additional Information Around Entering Origins]({{ site.baseurl }}/enterprise/server/#additionalinformationaroundenteringorigins) for more details.
```
ephox-mogran, line 31: > Other browser do not do this

I'm not sure that this is true any more. I think this was written a long time ago. We need to double-check this. If it's not true, we might as well delete the whole section.

ephox-mogran: - [ ] double-check this.

enterprise/server/configure.md
```diff
@@ -22,7 +22,7 @@ The Enhanced Media Embed server-side component require additional configuration,
 
 The `allowed-origins` element configures a list of **all** values that can be expected by the server-side components in a HTTP Origin header from your {{site.productname}} instances (see the [Mozilla Developer Network](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin) for more information on the HTTP Origin header). In short, you'll need to supply a list of all the URLs that your {{site.productname}} instances will be served from without the path information.
 
-> **Important:** When the services are matching against the `allowed-origins` list, the `scheme` (e.g. http) and the `port` (e.g. 800) are **ignored**.
+> **Important:** When the services are matching against the `allowed-origins` list, the `scheme` (e.g. http) and the `port` (e.g. 800) are **ignored** by default. Using the `ephox.allowed-origins.ignore-port` setting can customize this behaviour.
```
ephox-mogran, line 25: It won't change the scheme behaviour, but it will change the port behaviour. I'm not convinced that is adequately communicated by the sentence I have there, so I'm hoping that you change it.

enterprise/server/configure.md
```diff
@@ -138,6 +138,25 @@ ephox {
 }
 ```
 
+#### `allowed-origins.ignore-port` (optional)
+
+{% include misc/list_ignore_port_service_versions.md %}
+
+Disabling `ignore-port` means that the port of the origin **must match** the port of the origin specified in the `allowed-origins` list. This option is set to `true` by default, so the port is ignored.
```
JamesToohey, line 145:
```suggestion
Setting `ignore-port` to `false` means that the port of the origin **must match** the port of the origin specified in the `allowed-origins` list. This option is set to `true` by default, so the port is ignored.
```

enterprise/server/configure.md
```diff
@@ -138,6 +138,25 @@ ephox {
...
+
+Setting `ignore-port` to `true` will mean that the port is ignored when checking against the allowed origins. Setting it to `false` will mean that the port must match the origin when checking allowed-origins.
```
JamesToohey, line 147: We probably don't need this because it's state above

```suggestion
Setting `ignore-port` to `true` will mean that the port is ignored when checking against the allowed origins.
```

enterprise/server/troubleshoot.md
JamesToohey:
```suggestion
If the server is not running on a standard HTTP or HTTPS port (80 or 443) then Chrome will include the port in the Origin header that is sent to the server. Other browsers do not do this. If you have set `ignore-port` to `false`, then you will need to include the port in the `allowed-origins`. See [Additional Information Around Entering Origins]({{ site.baseurl }}/enterprise/server/#additionalinformationaroundenteringorigins) for more details.
```

lnewson (approved):

enterprise/server/configure.md
ephox-mogran: Yeah, it seems redundant to me, too. `same-origin` sort of had this pattern, but it was also a lot more complicated to explain.

enterprise/server/configure.md
ashfordneil:
```suggestion
> **Important:** When the services are matching against the `allowed-origins` list, the `scheme` (e.g. http) and the `port` (e.g. 800) are **ignored** by default. Customize this using [`ephox.allowed-origins.ignore-port`](#allowed-origins.ignore-port) option.
```

Maybe this?

ashfordneil (approved):

enterprise/server/configure.md
```diff
@@ -138,6 +138,25 @@ ephox {
 }
 ```
 
+#### `allowed-origins.ignore-port` (optional)
+
+{% include misc/list_ignore_port_service_versions.md %}
+
+Setting `ignore-port` to `false` means that the port of the origin **must match** the port of the origin specified in the `allowed-origins` list. This option is set to `true` by default, so the port is ignored.
+
+Setting `ignore-port` to `true` will mean that the port is ignored when checking against the allowed origins. Setting it to `false` will mean that the port must match the origin when checking allowed-origins.
```
tylerkelly13, line 147: Not sure if it's technically correct, but this might be a simpler description?
```suggestion
To enforce a port check for the service's `allowed-origins`, set `ignore-port` to `false`. When `ignore-port` is set to false, all `allowed-origins` **must** include the port of the origin.

This option is set to `true` by default, so ports specified in `allowed-origins` are ignored.
```

tylerkelly13 (commented):
we should probably make that description simpler, I've suggested an attempt, but it may need fixing up.

enterprise/server/configure.md
ephox-mogran: Seems correct.

tylerkelly13 (approved):

EkimChau (approved):

PR #2099, opened 2021-09-14
Add more notes

tylerkelly13:
Related Ticket: N/A

Description of Changes:
* add notes

Pre-checks:

Review:

### DIFF
--- a/release-notes/release-notes59.md
+++ b/release-notes/release-notes59.md
@@ -294,9 +294,9 @@ For information on:
 - The Spell Checker Pro plugin, see: [Spell Checker Pro plugin]({{site.baseurl}}/plugins/premium/tinymcespellchecker/).
 - Deploying the server-side components, see: [Server-side component installation]({{site.baseurl}}/enterprise/server/).
 
-### Stability update for the pre-configured server-side component Dockerfiles
+### Security and stability fixes for the pre-configured server-side component Dockerfiles
 
-The pre-configured Dockerfiles for containerizing the server-side components, also known as the Docker starter kits, have been updated to improve stability. Note that the resulting docker image will be larger as a result of these changes.
+The pre-configured Dockerfiles for containerizing the server-side components, also known as the Docker starter kits, have been updated to address various security issues and to improve stability. Note that the resulting docker image will be larger as a result of the stability changes.
 
 For information on deploying the server-side components using Docker, see: [Containerized service deployments]({{site.baseurl}}/enterprise/server/dockerservices/).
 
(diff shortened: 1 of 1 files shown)


release-notes/release-notes59.md
```diff
@@ -367,9 +367,9 @@ For information on:
 
 ## Security fixes
 
-{{site.productname}} 5.9 provides fixes for the following security issues:
+{{site.productname}} 5.9 provides fixes for the premium pre-configured Dockerfiles for containerizing the server-side components, also known as the Docker starter kits. These Dockerfiles been updated to address various security issues.
```
lnewson, line 370: This isn't correct to my knowledge, the WAR files are the parts that had the dependencies updated, not the docker files (though they do pull in the WAR files).

lnewson (changes requested):

release-notes/release-notes59.md
lnewson: Either way, these security updates apply to all server side items, not just the docker files.

ephox-mogran: The dockerfiles were updated as well. They now use jetty's image directly, rather than dealing with gpg keyservers themselves.

ephox-mogran: So the dockerfile updates were more about making them work more often (rather than need patching every time jetty changed contributors)

tylerkelly13: Updated the notes in commit https://github.com/tinymce/tinymce-docs/pull/2099/commits/a1b0741dd32a722366d5cb4f433953f210e1bbdd

release-notes/release-notes59.md
```diff
@@ -367,9 +371,7 @@ For information on:
 
 ## Security fixes
 
-{{site.productname}} 5.9 provides fixes for the following security issues:
-
-- changelog
+{{site.productname}} 5.9 provides fixes for the accompanying premium self-hosted server-side components. The `.war` files have been updated to address various security issues.
```
lnewson, line 374: Should this include a list of fixes, as this is something we normally include at a high level so customers were aware the a) severity of the security issue and b) roughly what they contained so they can determine the priority. The severity is really the key here, so we at minimum need to say if it's `Critical`, `High`, `Medium` or `Low` severity fixes.

- [ ] Ensure we at least include the severity of the security fixes.

lnewson (approved):

ephox-mogran (approved):

release-notes/release-notes59.md
ephox-mogran: This is my fault. I was trying to get out of providing exact CVEs because there were a few, but then there were more in between the releases after the first series of patches. If we need them, I can get them ... it's just a bit more involved because of the numerous patches between the 5.8 and 5.9 release. Some of which only related to things that were introduced after 5.8.

lnewson: I think it's fine if we don't provide the exact list, though that would be preferred, however we should at least mention what was the highest severity issue that was fixed here to give a better indication to customers as to how important it is they upgrade.

lnewson: For example, we could just say something like this (I have no idea if these levels are accurate):
```suggestion
{{site.productname}} 5.9 provides fixes for the accompanying premium self-hosted server-side components. The `.war` files have been updated to address various high and medium severity security issues.
```

ephox-mogran: Looks like high and medium is correct.

tylerkelly13: Thanks!

EkimChau (approved):

PR #2114, opened 2021-09-22
TINY-7894: Fixed incorrect live demo ids which caused the editor to not hide when swapping tabs

lnewson:
Related Ticket: TINY-7894

Description of Changes:

The live demos for the docs requires the demo name and the id used for the editor to be the same. This is a bit of a flaw and it doesn't work with demos that have multiple editors (e.g. inline, distraction free, RTC, etc...). As such we probably need to look at changing how that logic works but I want to talk to @tylerkelly13 about that first. So for now this fixes a number of issues reported as side notes in TINY-7894, which was caused by the editor not being hidden when swapping tabs.

Pre-checks:

Review:

### DIFF
--- a/_includes/live-demos/file-picker/index.js
+++ b/_includes/live-demos/file-picker/index.js
@@ -1,5 +1,5 @@
 tinymce.init({
-  selector: '#editor',
+  selector: '#file-picker',
   plugins: 'image code',
   toolbar: 'undo redo | link image | code',
   /* enable title field in the Image dialog*/
--- a/_includes/live-demos/file-picker/index.js
+++ b/_includes/live-demos/file-picker/index.js
@@ -1,5 +1,5 @@
 tinymce.init({
-  selector: '#file-picker',
+  selector: 'textarea#file-picker',
   plugins: 'image code',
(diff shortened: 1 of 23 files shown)


_includes/live-demos/file-picker/index.js
```diff
@@ -1,5 +1,5 @@
 tinymce.init({
-  selector: '#editor',
+  selector: '#file-picker',
```
ltrouton, line 2: For consistency?
```suggestion
  selector: 'textarea#file-picker',
```

ltrouton (approved):

_includes/live-demos/file-picker/index.js
lnewson: I also saw this, but didn't want to make this change as I assumed it wasn't done for a specific reason 🤷

tylerkelly13: > I assumed it wasn't done for a specific reason

probably not. it's pretty inconsistent. both are valid, but I prefer the explicit version.

tylerkelly13 (approved):

PR #2116, opened 2021-09-27
Fix minor inaccuracies in RTC snapshot handling

tjdett:
The snapshot callback is not executed at regular intervals, but instead after edits are made to the document.

Snapshots are also stored on the server, but only as a editor checkpoint so operations do not need to be replayed from the beginning of the document editing session. By replacing "regular" with "HTML", the statement is more correct. There may be some HTML snapshots currently being sent to the server, but as legacy behaviour that is likely to be removed. (In all cases, these snapshots are encrypted before being sent to the server, and not readable without the client.)

Pre-checks:

Review:

### DIFF
--- a/_includes/configuration/rtc_snapshot.md
+++ b/_includes/configuration/rtc_snapshot.md
@@ -4,7 +4,7 @@ Real-time collaboration integrations regularly store the content, eliminating th
 
 For any given document ID, the server guarantees the version number will only increase. It can be safely used for conflict resolution. For each document ID and version combination, the snapshot content is guaranteed to be identical.
 
-The snapshot callback will be executed shortly after edits are made with access to the serialized editor content. The content is retrieved through a `getContent` function to reduce CPU load if the callback decides to not use the editor content.
+The snapshot callback will be executed in response to local changes, with access to the serialized editor content. The content is retrieved through a `getContent` function to reduce CPU load if the callback decides to not use the editor content.
 
 {% if plugincode != "rtc" %}
 Required plugin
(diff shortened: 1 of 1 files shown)


_includes/configuration/rtc_snapshot.md
```diff
@@ -1,10 +1,10 @@
...
+Real-time collaboration integrations regularly store the content, eliminating the need for a save button. The {{site.productname}} RTC plugin provides a version number to assist with storing the HTML content snapshots. These snapshots are not stored on the RTC server and must be handled by the integrator.
 
 For any given document ID, the server guarantees the version number will only increase. It can be safely used for conflict resolution. For each document ID and version combination, the snapshot content is guaranteed to be identical.
 
-The snapshot callback will be executed at regular intervals with access to the serialized editor content. The content is retrieved through a `getContent` function to reduce CPU load if the callback decides to not use the editor content.
+The snapshot callback will be executed shortly after edits are made with access to the serialized editor content. The content is retrieved through a `getContent` function to reduce CPU load if the callback decides to not use the editor content.
```
lnewson, line 7: This isn't really any clearer either? Do this mean it makes a snapshot for each character, word, paragraph I add or what? I don't know enough about this to be able to make a suggestion sorry, but I think we need to be a little clearer here still.

If it's whenever a new undo level is created, maybe say that?

tylerkelly13: we don't really talk enough about undo levels...

TheSpyder: There’s no fixed timeline for snapshots, “in response to local changes” is not just a vague documentation phrasing - it’s the reality.

It depends on how fast the local user is making changes and their lag to the server. At a technical level the snapshot event is raised any time a local change is confirmed by the server with no pending local changes to send.

tylerkelly13:
```suggestion
The snapshot callback will be executed in response to local changes, with access to the serialized editor content. The content is retrieved through a `getContent` function to reduce CPU load if the callback decides to not use the editor content.
```
?

tylerkelly13: using Andy's suggestion

tylerkelly13 (approved):

lostkeys (approved):

TheSpyder (approved):

PR #2136, opened 2021-10-13
Workaround to fix syntax highlighting

tylerkelly13:
Co-authored-by: Lee Newson <lee.newson@tiny.cloud>

Related Ticket: N/A

Description of Changes:
* Workaround to fix syntax highlighting

Pre-checks:

Review:

### DIFF
--- a/_includes/template/bodycontents.html
+++ b/_includes/template/bodycontents.html
@@ -218,6 +218,17 @@
 </script>
 {% endif %}
 {% include_remote {{ site.origin }}/_docs/footer.html %}
+<script>
+  var elements = document.getElementsByClassName('highlighter-rouge');
+  for (var i = 0; i < elements.length; i++) {
+      var element = elements[i];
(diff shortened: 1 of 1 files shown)


lnewson:
For context, Tyler and I found out last night that the docs has quite a strange setup for code highlighting... first it must use a valid language per `rouge` (a ruby package) and that will render the initial markup as something like this:
```
<div class="language-js highligher-rouge">
  <div class="highlight">
    <pre class="prettyprint>
      <code>
      ...
      </code>
    </pre>
  </div>
</div>
```
From that, a second library (code-prettify) from the main site is used to blow away all markup inside the code and tries to a detect the language and highlight accordingly. Effectively this means the language we tell it to use in the docs is never actually respected for what's shown. So the core of this change is to find everything highlighted by `rouge` and move the language class to the `pre` element where `code-prettify` can process it and highlight correctly.

PR #2225, opened 2022-02-14
Feature/6.0 api docs

EkimChau:
Related Ticket: 

Description of Changes:
* Placeholder text

Pre-checks:

Review:


spocke (changes requested):

_scripts/api-reference.sh
```diff
@@ -2,21 +2,33 @@
 
 set -e
 
-API_VERSION="$(cat .api-version)"
+API_VERSION="develop"
 TARBALL_URL="https://codeload.github.com/tinymce/tinymce/tar.gz/$API_VERSION"
```
TheSpyder, line 6: Interesting. This suggests we don't need published releases to compile API docs against, as long as the API version tag contains code with the correct docs.

TheSpyder (approved):

_scripts/api-reference.sh
EkimChau: There's also a build from local source flavour _scripts/api-reference-local.sh thats wired into yarn build-local ../path/to/local/tinymce

TheSpyder: yeah I saw that, I meant in reference to a conversation yesterday about when source code had to be changed for API docs.

tiny-james (approved):

_scripts/api-reference.sh
```diff
@@ -2,21 +2,33 @@
 
 set -e
 
-API_VERSION="$(cat .api-version)"
+API_VERSION="develop"
```
lnewson, line 5: @EkimChau heads up we'll need to make sure this change is reverted when we release 6.0, otherwise it'll be building the incorrect API docs.

PR #2228, opened 2022-02-15
WSRF- 609 - Add demo support

tinyamit:
Related Ticket: WSRF-609

Description of Changes:
* Updated the Antora Docs to support the Demos

Pre-checks:
- [n/a] `_data/nav.yml` has been updated (if applicable)

Review:


modules/ROOT/pages/demo/local-upload.adoc
```diff
@@ -11,7 +12,40 @@ image::{baseurl}/images/demo_local-upload_01.png[Upload tab]
...
+----
+include::example$live-demos/local-upload/index.js[]
+----
+--
+Edit on Codepen::
++
+--
+Codepen item goes here
+--
+====
```
lnewson, line 48: This is pretty horrific to try and remember the exact format needed for this... it should be done via a macro like it was previously, as examples should be easy to add.

lnewson (changes requested):
I'm sorry, but this new format is way to verbose/hard to use as maintainers for the docs. We need something simple like what was there previously.

lnewson:
Another issue I see with this, is it removes the ability to use variables within the demos which means everything needs to be duplicated and makes this harder to maintain. So it's not the path we should be taking here, as this system is supposed to make things better/easier to maintain.

modules/ROOT/pages/demo/local-upload.adoc
tinyamit: Antora /ASCII doc does not give many options to run JS , the only way I got it is to use pass through block. Otherwise other way around will be dumping all these file to the UI and perform some macro in. But certainly that defies the purpose of have version controlled branches. 
This solution was originally presented by the creator of Antora and is also used by couch base too. There might be a way around this but it may require deeper digging and understanding working of Antora. 

Being said, I am all ears. If you have some better solution I will be happy to listen.

tinyamit:
> 

I understand that, I have a way around that. But there will still be limitations in regards to what props can be passed. Its usually strings and boolean which can be used as attributes for doc

modules/ROOT/pages/demo/local-upload.adoc
lnewson: As I said and as proved by the previous version, this massive block can be generated by macros. It also allows embedding the JS or creating scripts like this automatically so that writers don't need to manually do it all the time.

tinyamit: actually yes, I will change that soon.

lnewson: On that note, what was the issue with the thing that Tyler and I had whipped up as a PoC? The Jira mentions nothing and I can't see why it needs to be replaced.

tinyamit:
@lnewson have a look now ?

lnewson:
> @lnewson have a look now ?

That's better, but sorry @tinyamit it's still not going to work. The demos need to be dynamic so it needs to be something similar to what was there originally. As as example, sometimes it needs to load from `example.js` for the content to show instead of `index.js` and other times it shouldn't include the CodePen link (e.g. if an example.js file is included there should be no codepen link). I also suspect this wouldn't actually work, as nothing appears to be including the relevant TinyMCE or Tiny Drive scripts from cloud (I can't get the demo working to test). So overall this solution does not appear viable and doesn't meet the requirements needed for documentation authoring and demos.

This really should have been discussed with us before this task was undertaken to be honest, given we are the users/stakeholders when it comes to authoring. You still also haven't answered my previous question:
> On that note, what was the issue with the thing that Tyler and I had whipped up as a PoC? The Jira mentions nothing and I can't see why it needs to be replaced. It was definitely a better/more maintainable solution than manually having to write this out

EkimChau (approved):

lnewson:
I'm going to close this because this isn't an approach we're taking at the moment. We can come back to this later if needed.

PR #2235, opened 2022-02-25
DOC-1532: Fixed markdown -> asciidoc conversion issues for integrations

lnewson:
Related Ticket: DOC-1532

Description of Changes:
* Replace all liquid logic with asciidoc ifeval, etc... logic instead
* Fixed other conversion issues caused by the liquid syntax that prevented it being converted automatically
* Fixed conversion issues that caused numbering issues in lists
* Fixed conversion issues with admonitions (these were converted to blockquotes since markdown doesn't have this concept)
* Fixed xref conversion issues (basically replace the legacy `<<id, title>>` format with `xref:id[title]`)

**Note:** I'd suggest running this locally instead of trying to review the actual markup here.

Pre-checks:

Review:


EkimChau (approved):

TheSpyder (approved):

lnewson:
As discussed at standup we're going to skip reviewing every change here as it's holding us up too much and there's a lot to fix. Instead we'll review the content as a whole at the end and then merge from this branch. As such, I'm going to convert this to a draft for now.

PR #2236, opened 2022-02-25
Feature/tiny 8441

EkimChau:
Related Ticket: 

Description of Changes:
* Updated docs repo to fill missing landing pages
* Updated minor details to reflect 6.0 changes

Pre-checks:
- [ ] Branch prefixed with `feature/` or `hotfix/`
- [ ] `_data/nav.yml` has been updated (if applicable)

Review:

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -14,6 +14,9 @@
 **** xref:laravel-tiny-cloud.adoc[Laravel]
 **** xref:rails-cloud.adoc[Ruby on Rails]
 *** Self-hosted
+**** JavaScript
+***** xref:javascript-pm.adoc[Using a package manager]
+***** xref:javascript-zip.adoc[Using a .zip package]
 **** React
 ***** xref:react-pm.adoc[Using a package manager]
 ***** xref:react-zip.adoc[Using a .zip package]
@@ -48,6 +51,7 @@
 **** xref:dotnet-projects.adoc[.NET projects]
 **** xref:bower-projects.adoc[Bower projects]
 **** xref:zip-install.adoc[TinyMCE .zip deployments]
+** xref:integrations.adoc[TinyMCE Integrations]
 ** xref:upgrading.adoc[Upgrading TinyMCE]
 * How-to guides
 ** Learn the basics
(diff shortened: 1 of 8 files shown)


modules/ROOT/nav.adoc
```diff
@@ -14,6 +14,9 @@
 **** xref:laravel-tiny-cloud.adoc[Laravel]
 **** xref:rails-cloud.adoc[Ruby on Rails]
 *** Self-hosted
+**** JavaScript
```
lnewson, line 17: This should not be here... these pages already exist as `zip-install.adoc` and `npm-projects.adoc`

modules/ROOT/nav.adoc
```diff
@@ -48,6 +51,7 @@
 **** xref:dotnet-projects.adoc[.NET projects]
 **** xref:bower-projects.adoc[Bower projects]
 **** xref:zip-install.adoc[TinyMCE .zip deployments]
+** xref:integrations.adoc[TinyMCE Integrations]
```
lnewson, line 54: This was specifically removed/moved to the installation section from what I recall as this was part of Content Strategy changes for 6.0. I don't think we should have this, but we should check with Brian.

lnewson (changes requested):
This looks like it's still a WIP? If so this should be a draft PR so I'm going to convert it to that.

modules/ROOT/nav.adoc
EkimChau: Yeah, there are parts of the docs that link to see integrations, which is very generic, so I put this placeholder in.  It would be just a simple integrations index that filters out to specific integrations

modules/ROOT/nav.adoc
EkimChau: Right, okay we can leave this here as a placeholder to fix

lnewson: There's nothing to fix here... those are already there you just need to delete/revert these changes. Also remember that most of these are JavaScript, so having a heading like this would only be confusing.

EkimChau: This will also need to be reverted in the tinymce source code aswell, will follow that up later, leaving as is for now

PR #2254, opened 2022-03-07
DOC-987/DOC-1530/DOC-1543: Document format, command, schema changes

ltrouton:
Related Ticket: DOC-987, DOC-1530, DOC-1543

Description of Changes:
* DOC-987: Document alignleft and alignright format change for tables
* DOC-1530: Document mceAddEditor and mceToggleEditor command value change
* DOC-1543: Update valid_elements documentation to replace : with ~

Pre-checks:
- [X] Branch prefixed with `feature/` or `hotfix/`
- [X] `_data/nav.yml` has been updated (if applicable)
- [X] ~~(New product features only) Release Note added~~

Review:

### DIFF
--- a/modules/ROOT/partials/configuration/extended_valid_elements.adoc
+++ b/modules/ROOT/partials/configuration/extended_valid_elements.adoc
@@ -7,7 +7,7 @@ When adding a new attribute by specifying an existing element rule (e.g. `+img+`
 
 The below example replaces the current `+img+` rule (including the global element rules).
 
-Type: `+String+`
+Type : `+String+`
 
 === Example: Using `+extended_valid_elements+`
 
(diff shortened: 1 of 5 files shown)


modules/ROOT/partials/configuration/extended_valid_elements.adoc
```diff
@@ -7,7 +7,7 @@ When adding a new attribute by specifying an existing element rule (e.g. `+img+`
 
 The below example replaces the current `+img+` rule (including the global element rules).
 
-Type : `+String+`
```
lnewson, line 10: This and other changes like this shouldn't have been done, as everywhere now uses this style and it's what Tyler changed it to. I believe it was meant to be definition lists instead (e.g. `::`) but I'm personally not a fan of that for these.

Either way, this is something we need to change en masse if we're going to and shouldn't be done for these couple of specific cases. It probably needs a discussion or chat with Brian to figure out what's best and then should be done via a separate PR.

- [ ] Revert removing these spaces for now

lnewson (approved):

modules/ROOT/partials/configuration/extended_valid_elements.adoc
ltrouton: Ah, I didn't release that. Good to know. Hopefully should be good now

EkimChau (approved):

PR #2276, opened 2022-03-14
DOC-1587: Add beta prefix to docs title

EkimChau:
Related Ticket: 

Description of Changes:
* Add [Beta] to the docs title, to make it clear this is not production grade docs

Screen Shot 2022-03-14 at 4 11 59 pm

Pre-checks:

Review:

### DIFF
--- a/antora.yml
+++ b/antora.yml
@@ -1,5 +1,5 @@
 name: tinymce
-title: TinyMCE Documentation
+title: TinyMCE Documentation [BETA]
 version: '6'
 asciidoc:
   attributes:


antora.yml
```diff
@@ -1,5 +1,5 @@
 name: tinymce
-title: TinyMCE Documentation
+title: TinyMCE Documentation [BETA]
```
lnewson, line 2: This won't work, as this would mark all documentation as "beta" (e.g. for 5.x as well). You'll need to do it on the version instead or find another suitable variant.

lnewson (changes requested):

antora.yml
EkimChau: as this would mark all documentation as "beta"

how so? I could add [BETA 6.0]

lnewson: Antora groups things by this title/product name (take a look at staging down in the bottom right), so this would either end up showing as a separate product or would change all other products. So we need to change something that's going to be specific to 6.0 instead. I guess if you're happy with it showing as a separate product then this would be okay, but that seemed like a less than ideal solution.

lnewson: image
Screenshot as that might be easier.

lnewson: Discussed this more in a zoom call this morning. For now we're only going to release the TinyMCE 6 docs and by the time we need to add others we should no longer need this. If we do then we'll re-evaluate then.

lnewson (approved):

PR #2279, opened 2022-03-14
Feature-6-DOC-1584 - Prepare TinyMCE 6.0 docs for automated deployments to staging and production

tinyamit:
Related Ticket: https://ephocks.atlassian.net/browse/DOC-1584

Description of Changes:
* Setup pipeline for staging. dev and release branches for tinymce 6 docs build and deploy.
* Setup was credential and theme buckets
* configure package.json
* currently targets:
- `feature-6/**` for feature branches
- `staging/TinyMCE_6.0` for staging branch
- `release/TinyMCE_6.0` for release branch

Pre-checks:
- [n/a] `_data/nav.yml` has been updated (if applicable)
- [n/a] (New product features only) Release Note added

Review:

### DIFF
--- a/.github/workflows/feature_6_docs.yml
+++ b/.github/workflows/feature_6_docs.yml
@@ -28,7 +28,7 @@ jobs:
         node-version: ${{ matrix.node-version }}
 
     - name: (Feature - Docs 6.0) Install dependencies
-      run: yarn install
+      run: yarn install -g
 
     - name: (Feature - Docs 6.0) Build API References
       run: yarn build
--- a/.github/workflows/feature_6_docs.yml
+++ b/.github/workflows/feature_6_docs.yml
@@ -1,0 +1,60 @@
+name: Develop > Tiny docs 6.0
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/**'
+      - '*'
+    branches:
+      - 'feature-6/**'
+
+jobs:
+  build:
+    name: Build Docs and Deploy
+    runs-on: ubuntu-18.04
+
+    strategy:
+      matrix:
+        node-version: [14]
+
+    steps:
+    - uses: actions/checkout@v1
+
+    - name: Use Node.js
+      uses: actions/setup-node@v2
+      with:
+        cache: 'yarn'
+        node-version: ${{ matrix.node-version }}
+
+    - name: (Feature - Docs 6.0) Install dependencies
+      run: yarn install
+
+    - name: (Feature - Docs 6.0) Build API References
+      run: yarn build
+
+    - name: (Feature - Docs 6.0) Build Website
+      run: antora ./antora-playbook-staging.yml
+
+    - name: (deploy) Prepare for branch deployments
+      shell: bash
+      run: |
+        printf "User-agent: * \nDisallow: /\n" > ./build/site/robots.txt
+        echo $(node -e 'console.log(`s3://docs-${process.env.GITHUB_REF.replace("refs/heads/","").replace(/[/\s]/g, "-").replace(/[^0-9a-zA-Z-]/g, "").replace(/-$/g, "").toLowerCase()}.staging.tiny.cloud`)') > S3_BUCKET
+        printf "S3_BUCKET: ${S3_BUCKET} (${#S3_BUCKET})\n"
+
+        if [ ${#S3_BUCKET} -lt 3 ]
+        then
+          echo "WARNING! S3_BUCKET name is too short"
+        fi
+
+        if [ ${#S3_BUCKET} -gt 63 ]
+        then
+          echo "WARNING! S3_BUCKET name is too long"
+        fi
+
+    - name: (deploy) Upload website to S3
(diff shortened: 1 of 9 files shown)


.github/workflows/staging_6_docs.yml
```diff
@@ -0,0 +1,52 @@
...
+      run: sudo yarn global add @antora/cli@3.0.1
+
+    - name: (Staging - Docs 6.0) Install dependencies
+      run: yarn install -g
+
+    - name: (Staging - Docs 6.0) Build API References
+      run: yarn build
+
+    - name: (Staging - Docs 6.0) Build Website
+      run: antora ./antora-playbook-staging.yml
```
EkimChau, line 40: we should add a robots no-index rule on staging

EkimChau (approved):

.github/workflows/feature_6_docs.yml
```diff
@@ -0,0 +1,63 @@
+name: Develop > Tiny docs 6.0
+
+on:
+  workflow_dispatch:
+  push:
+    paths:
+      - '.github/workflows/**'
+      - '*'
+    branches:
+      - 'feature-6/**'
```
lnewson, line 10: This goes against the branching strategy being used here and elsewhere... so I don't think adding `6` like this is a good idea. I'm not sure what's better, but I'd prefer either `feature/6/<JIRA>`, `feature/<JIRA>_6.x` or even better just make it work for any `feature/<JIRA>` branch. It might need a wider discussion to be honest.

.github/workflows/feature_6_docs.yml
```diff
@@ -0,0 +1,63 @@
...
+      uses: actions/setup-node@v2
+      with:
+        cache: 'yarn'
+        node-version: ${{ matrix.node-version }}
+
+    - name: Install Antora@3.0.1
+      run: sudo yarn global add @antora/cli@3.0.1
+
+    - name: (Feature - Docs 6.0) Install dependencies
+      run: yarn install -g
```
lnewson, line 34: Why are we doing this? We have a package.json and lockfile for a reason... we should just be able to run `yarn install` instead here.

.github/workflows/feature_6_docs.yml
```diff
@@ -0,0 +1,63 @@
...
+
+    - name: (Feature - Docs 6.0) Build API References
+      run: yarn build
+
+    - name: (Feature - Docs 6.0) Build Website
+      run: antora ./antora-playbook-staging.yml
```
lnewson, line 40: We don't need to use the antora global here, as we should be using the one installed by yarn. As such, I'd suggest doing something like this instead:
```suggestion
      run: yarn antora ./antora-playbook-staging.yml
```

lnewson: Also should this be using the `dev` playbook instead of the `staging` one?

antora-playbook-dev.yml
```diff
@@ -0,0 +1,15 @@
+site:
+  title: TinyMCE Documentation (Dev)
+  url: https://tiny.cloud/docs
```
lnewson, line 3: Would this be better off using the staging site as well?
```suggestion
  url: https://staging.tiny.cloud/docs
```

.github/workflows/feature_6_docs.yml
EkimChau: currently feature/ is mapped to the legacy 5x docs
feature-6/ new 6.0
feature-5/ new 5.0

package.json
```diff
@@ -27,8 +27,8 @@
     "ext": "*"
   },
   "devDependencies": {
-    "@antora/cli": "^2.3.4",
-    "@antora/site-generator-default": "^2.3.4",
+    "@antora/cli": "^3.0.1",
```
lnewson, line 30: This was already done in the release branch, so does this need the latest changes merged into this `staging/TinyMCE_6.0` branch?

.github/workflows/feature_6_docs.yml
lnewson: Yeah, as I said that's only going to cause issues going forward. We need to keep it `feature/` as the prefix always. If the sole reason we can't do that is because of the old system, then why do we even need the old branches as we won't be using that anymore?

Also this is another thing that really should have been discussed with us before making a decision yourselves 😞

.github/workflows/feature_6_docs.yml
tinyamit: My bad, I didn't clean it well

.github/workflows/feature_6_docs.yml
tinyamit: absolutely !

package.json
tinyamit: yep, we have to do it in release too.

lnewson: What do you mean? As I said it's already done in the release branch, which means this is behind the release branch.

.github/workflows/feature_6_docs.yml
tinyamit: winner => feature/6/<JIRA>

PR #2292, opened 2022-03-21
DOC-1590: Updated the snippets and demos in the How-to guides section

lnewson:
Related Ticket: DOC-1590

Description of Changes:
* Updated the demos/snippets to use ES6 of above syntax
* Moved the css/js example files to the `examples` directory from the `partials` directory (since that's what the dir exists for)
* Fixed various whitespace issues in the demos
* Updated the demos to work with TinyMCE 6 (the `images_upload_handler` change does overlap a little with DOC-1572 unfortunately)
* Fixed other minor miscellaneous issues found while going through each section
* Made the server side conf "language" consistent as it was a mix of `conf` and `properties`, with `conf` not being a language that is compatible with the highlighter.

Pre-checks:

Review:


modules/ROOT/partials/configuration/defaultmenuitems.adoc
```diff
@@ -3,13 +3,13 @@
...
-    insert: { title: 'Insert', items: 'image link media template codesample inserttable | charmap emoticons hr | pagebreak nonbreaking anchor | insertdatetime' },
-    format: { title: 'Format', items: 'bold italic underline strikethrough superscript subscript codeformat | styles blocks fontfamily fontsize align lineheight | forecolor backcolor | removeformat' },
-    tools: { title: 'Tools', items: 'spellchecker spellcheckerlanguage | code wordcount' },
-    table: { title: 'Table', items: 'inserttable | cell row column | tableprops deletetable' },
+    file: { title: 'File', items: 'newdocument restoredraft | preview | export print | deleteallconversations' },
+    edit: { title: 'Edit', items: 'undo redo | cut copy paste pastetext | selectall | searchreplace' },
+    view: { title: 'View', items: 'code | visualaid visualchars visualblocks | spellchecker | preview fullscreen | showcomments' },
+    insert: { title: 'Insert', items: 'image link media addcomment pageembed template codesample inserttable | charmap emoticons hr | pagebreak nonbreaking anchor tableofcontents | insertdatetime' },
```
EkimChau, line 9: Do we need a followup Jira to remove tableofcontents from the defaults?

lnewson: No because it is always in the defaults. It's what has to be done to get the items included in the menus, see: https://github.com/tinymce/tinymce/blob/236891ffcdf2fd1b698d020e67c31838fb110019/modules/tinymce/src/themes/silver/main/ts/ui/menus/menubar/Integration.ts#L25. You'll notice multiple other premium plugins are listed as well. To solve the part where they don't have to be listed in core is an entirely separate thing that should be discussed elsewhere if we want to go down that path.

EkimChau (approved):

modules/ROOT/partials/configuration/defaultmenuitems.adoc
lnewson: I should add if we just don't want premium plugins documented that's up to you 🤷

metricjs (approved):

PR #2302, opened 2022-03-24
DOC-1592: Updated snippets in the Content section for 6.0

tiny-ben-tran:
Related Ticket:  DOC-1592

Description of Changes:
* Updated snippets/demos to ES6
* Removed `gecko_spellcheck` section which is no longer supported in 6.0
* Changed the wording of `paste_data_images` option
* Removed the name of options where it is repeated in the beginning of the description, so it will be consistent with others
Pre-checks:
- [x] ~`_data/nav.yml` has been updated (if applicable)~

Review:


modules/ROOT/pages/spell-checking.adoc
```diff
@@ -1,7 +1,7 @@
 = Check spelling in TinyMCE
 
 :title_nav: Spell checking
-:keywords: spell checker spelling browser_spellcheck gecko_spellcheck {productname} provides several options to bring spell checking capabilities to your users.
+:keywords: spell checker spelling browser_spellcheck {productname} provides several options to bring spell checking capabilities to your users.
```
lnewson, line 4: Not related to this change but there's actually a conversion issue here by the looks, as most of this isn't keywords.

```suggestion
:keywords: spell checker spelling browser_spellcheck

{productname} provides several options to bring spell checking capabilities to your users.
```

modules/ROOT/partials/configuration/convert_fonts_to_spans.adoc
```diff
@@ -1,7 +1,7 @@
 [[convert_fonts_to_spans]]
 == convert_fonts_to_spans
 
-include::partial$DEPRECATED/generic-5_10.adoc[]
+IMPORTANT: This option has been deprecated in {productname} 5.10 and may be removed in {productname} 6.0.
```
lnewson, line 4: This isn't correct as this is the 6.0 docs... also we should update the original admonition not this. We probably want to say something like this:
```
IMPORTANT: This option has been deprecated in {productname} 5.10 and may be removed in a future major {productname} release.
```

- [ ] Revert and fix original admonition instead

modules/ROOT/partials/configuration/forced_root_block.adoc
```diff
@@ -18,6 +18,6 @@ Type : `+String+`
 ----
 tinymce.init({
   selector: 'textarea',  // change this value according to your HTML
-  forced_root_block : 'p'
+  forced_root_block: 'p'
```
lnewson, line 21: `p` is the default here so we should use something else. We also should add this between the type and example:
```
Default Value: `+'p'+`
```

modules/ROOT/partials/configuration/paste_data_images.adoc
```diff
@@ -5,11 +5,11 @@ This option specifies whether `+data:url+` images (inline images) should be remo
 
 Setting `+paste_data_images+` to `+true+` will allow the pasted images, while setting it to `+false+` will disallow pasted images.
 
-For example, Firefox enables you to paste images directly into any `+contentEditable+` field. This is normally not something people want, so this option is `+false+` by default. For example, a 600kb embedded image would block page loads and prevents it from being cached on multiple pages.
+For example, Firefox enables you to paste images directly into any `+contentEditable+` field. A 600kb embedded image would block page loads and prevents it from being cached on multiple pages. Setting this option to `+false+` would be prevent such case.
```
lnewson, line 8: This paragraph makes no sense... can you please have another go at rewording this or ask for help? We maybe event should just largely remove all this and just mention that it can be set to false to disable the behaviour.

- [ ] Reword this so it makes sense

modules/ROOT/partials/configuration/urlconverter_callback.adoc
```diff
@@ -5,10 +5,10 @@ This option enables you to add your own URL converter logic. This option should
...
 The format of the converter function is: `+URLConverter(url, node, on_save)+`, where:
 
-* url = the URL string to convert,
-* node = the element that contains the URL that is to be converted (this parameter may be set to null if there is no element for the URL), and
-* on_save is always set to true.
-* name = the attribute name that is being set.
+* `+url+`: The URL string to convert.
+* `+node+`: The element that contains the URL that is to be converted (this parameter may be set to null if there is no element for the URL).
+* `+on_save+`: Always set to true.
+* `+name+`: The attribute name that is being set.
```
lnewson, line 11:
```suggestion
* `+url+` is the URL string to convert.
* `+node+` is the element that contains the URL that is to be converted (this parameter may be set to null if there is no element for the URL).
* `+on_save+` is always set to true.
* `+name+` is the attribute name that is being set.
```
List points should continue on from the sentence above, so making it start new like this didn't make sense so I've changed it to keep the flow.

lnewson (approved):

modules/ROOT/partials/configuration/paste_data_images.adoc
metricjs:
```suggestion
For example, Firefox enables you to paste images directly into any `+contentEditable+` field. This functionality can cause problems for users, such as when attempting to paste an embedded image that is too large. This would block page loads and prevent it from being cached on multiple pages. Setting this option to `+false+` prevents such issues.
```
I made an attempt, but I'm not sure that quite makes sense yet either. 

I do like explaining why this defaults to false, but I definitely think we can generalise the example a bit.

lnewson: @metricjs it doesn't default to `false` though?

modules/ROOT/pages/spell-checking.adoc
EkimChau: looks like the old website omitted the meta description, could we also add one now so we have both keyword and description meta data?

:description: {productname} provides several options to bring spell checking capabilities to your users.

EkimChau (approved):

modules/ROOT/partials/configuration/forced_root_block.adoc
```diff
@@ -12,12 +12,14 @@ WARNING: Not using `+p+` elements as the root block will impair the functionalit
 
 Type : `+String+`
 
+Default Value: `+p+`
```
lnewson, line 15:
```suggestion
Default Value: `+'p'+`
```
As this is a string it needs quotes.

modules/ROOT/partials/configuration/line_height_formats.adoc
```diff
@@ -12,8 +12,8 @@ Default Value : `+'1 1.1 1.2 1.3 1.4 1.5 2'+`
 [source,js]
 ----
 tinymce.init({
-  selector: 'textarea', // change this value according to your HTML
-  toolbar: 'lineheight',
-  line_height_formats: '1 1.1 1.2 1.3 1.4 1.5 2'
+    selector: 'textarea', // change this value according to your HTML
+    toolbar: 'lineheight',
+    line_height_formats: '1 1.2 1.4 1.6 2'
```
ltrouton, line 17: Looks like this has incorrect indentation now?

ltrouton (approved):

modules/ROOT/partials/configuration/paste_data_images.adoc
tiny-ben-tran: The `paste_data_image` section reads better without the example paragraph. Beside, it doesn't add much value

modules/ROOT/partials/configuration/line_height_formats.adoc
tiny-ben-tran: I blame merge conflicts

modules/ROOT/partials/configuration/paste_data_images.adoc
metricjs: Agreed, let's remove the example

lnewson (approved):

PR #2311, opened 2022-03-25
DOC-1550: Add deprecation message to SetContent event

tiny-ben-tran:
Related Ticket: DOC-1550

Pre-checks:
- [x] ~`_data/nav.yml` has been updated (if applicable)~
- [X] ~(New product features only) Release Note added~

Review:

### DIFF
--- a/modules/ROOT/pages/events.adoc
+++ b/modules/ROOT/pages/events.adoc
@@ -118,7 +118,7 @@ The following events are provided by the {productname} editor.
 |BeforeGetContent |`+{ format: string, source_view?: boolean, selection?: boolean, save?: boolean }+` |Fired before the content is serialized from the editor.
 |GetContent |`+{ content: string, format: string, source_view?: boolean, selection?: boolean, save?: boolean }+` |Fired after the content is serialized from the editor.
 |BeforeSetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired before the content is parsed and rendered in the editor.
-|SetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired after the content is parsed and rendered in the editor.
+|SetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired after the content is parsed and rendered in the editor. *Note*: `+content+` argument has been deprecated and will be removed in the next major release.
 |LoadContent |N/A |Fired after the initial content has loaded into the editor.
 |PreviewFormats |N/A |Fired before a formats CSS is generated when the format is being previewed in the editor.
 |AfterPreviewFormats |N/A |Fired after a formats CSS is generated when the format is being previewed in the editor.
--- a/modules/ROOT/pages/events.adoc
+++ b/modules/ROOT/pages/events.adoc
@@ -118,7 +118,7 @@ The following events are provided by the {productname} editor.
 |BeforeGetContent |`+{ format: string, source_view?: boolean, selection?: boolean, save?: boolean }+` |Fired before the content is serialized from the editor.
 |GetContent |`+{ content: string, format: string, source_view?: boolean, selection?: boolean, save?: boolean }+` |Fired after the content is serialized from the editor.
 |BeforeSetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired before the content is parsed and rendered in the editor.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/events.adoc
```diff
@@ -118,7 +118,7 @@ The following events are provided by the {productname} editor.
 |BeforeGetContent |`+{ format: string, source_view?: boolean, selection?: boolean, save?: boolean }+` |Fired before the content is serialized from the editor.
 |GetContent |`+{ content: string, format: string, source_view?: boolean, selection?: boolean, save?: boolean }+` |Fired after the content is serialized from the editor.
 |BeforeSetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired before the content is parsed and rendered in the editor.
-|SetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired after the content is parsed and rendered in the editor.
+|SetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired after the content is parsed and rendered in the editor. *Note*: `+content+` argument has been deprecated and will be removed in the next major release.
```
tiny-ben-tran, line 121: `NOTE:` doesn't work in table

lnewson:
```suggestion
|SetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired after the content is parsed and rendered in the editor. NOTE: The `+content+` property has been deprecated and will be removed in the next major release.
```
Note sure what's best here, but I think it'd be better to keep the note markup consistent. As for the other changes, minor grammar fix and this isn't a function it's an object so `argument` isn't quite the right term.

Fore reference, you can get admonitions to render in tables by tweaking how the column is rendered: https://docs.asciidoctor.org/asciidoc/latest/tables/format-column-content/. However, in saying that it's probably not a great idea as admonitions are block content.

lnewson (approved):

modules/ROOT/pages/events.adoc
tiny-james: What is the reason for removing it? It seems odd that the SetContent event would not have the content being set?

lnewson: The `SetContent` event is fired after it's been inserted so you can't do anything with it... it's also quite expensive to compute that for RTC given its model based. The `BeforeSetContent` event still has the data available. https://github.com/tinymce/tinymce/pull/7603 and https://github.com/tinymce/tinymce/pull/7571#discussion_r798602309 have more details.

Anyways, if you want more info or want to discuss we're probably better doing that via slack or similar.

tiny-james: So the workaround is to switch to `BeforeSetContent` or to use `editor.getContent()` in the event handler?
I wonder if we could provide a method to produce the content that was set, like we have done in [rtc_snapshot](https://www.tiny.cloud/docs/rtc/configuration/rtc-options-optional/#rtc_snapshot).

lnewson: Yeah that's correct. As for a method, that might be viable as well if we feel it's needed, though I personally don't think it's needed for this specific case given you already have access to an editor if you're listening to the `SetContent` event 🤷  That said, it could be useful all-round though, as the content is often expensive in varying cases so making it lazy would help with that.

EkimChau (approved):

modules/ROOT/pages/events.adoc
```diff
@@ -118,7 +118,7 @@ The following events are provided by the {productname} editor.
 |BeforeGetContent |`+{ format: string, source_view?: boolean, selection?: boolean, save?: boolean }+` |Fired before the content is serialized from the editor.
 |GetContent |`+{ content: string, format: string, source_view?: boolean, selection?: boolean, save?: boolean }+` |Fired after the content is serialized from the editor.
 |BeforeSetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired before the content is parsed and rendered in the editor.
-|SetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired after the content is parsed and rendered in the editor.
+|SetContent|`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired after the content is parsed and rendered in the editor. NOTE: The `+content+` property has been deprecated and will be removed in the next major release.
```
lnewson, line 121:
```suggestion
|SetContent |`+{ content: string, format?: string, paste?: boolean, selection?: boolean }+` |Fired after the content is parsed and rendered in the editor. NOTE: The `+content+` property has been deprecated and will be removed in the next major release.
```
You accidentally deleted the space here in the last commit by the looks.

TheSpyder (approved):

PR #2332, opened 2022-03-29
DOC-1563: Renamed files and anchors for SEO and other purposes

lnewson:
Related Ticket: DOC-1563

Description of Changes:
* Changed how Antora/AsciiDoc generates ids to use hyphens and to exclude the underscore prefix
* Renamed various files to remove the underscore in the name where it wasn't required
  * Note: No pages were renamed in the end as they'd all been done, so only partials, etc... were renamed. Additionally the 6.0 release note pages still use underscores but whether or not they'll be kept is still to be determined so we'll need to come back to that. 
* Renamed anchors for heading to use hyphens instead of underscores or in many cases where no separator was used at all
* Added legacy anchors so links to specific anchors in TinyMCE don't break

Pre-checks:

Review:


modules/ROOT/pages/content-behavior-options.adoc
```diff
@@ -11,6 +11,8 @@ include::partial$configuration/draggable_modal.adoc[]
 
 include::partial$configuration/keep_styles.adoc[]
 
+include::partial$configuration/editable_class.adoc[]
```
lnewson, line 14: I intentionally moved this as the links to these sections were different and it shouldn't have been. As such it was a bug that they were in different pages.

thedrjim (approved):

metricjs (approved):

PR #2362, opened 2022-04-01
DOC-1627: RTC does not support Input Method Editors

spocke:
Related Ticket: 

Description of Changes:
* Not sure how deep we should explain IMEs but people who care about that would know that it is. I don't want to list all the languages that uses IMEs.

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/pages/rtc-supported-functionality.adoc
+++ b/modules/ROOT/pages/rtc-supported-functionality.adoc
@@ -20,11 +20,11 @@ Real-Time Collaboration is currently not supported on mobile devices.
 
 == Input support
 
-The Real-Time Collaboration (RTC) plugin currently doesn't support Input Method Editor (IME) as a input method.
+Real-Time Collaboration does not currently support using an Input Method Editor (IME).
 
 == TinyMCE features (plugins) support when using RTC
 
-Real-Time Collaboration does not currently support using an Input Method Editor (IME).
+The Real-Time Collaboration (RTC) plugin supports a subset of the features that {productname} provides.
 
 WARNING: Enabling plugins not shown here is not supported and may lead to unexpected behavior.
 
--- a/modules/ROOT/pages/rtc-supported-functionality.adoc
+++ b/modules/ROOT/pages/rtc-supported-functionality.adoc
@@ -24,7 +24,7 @@ The Real-Time Collaboration (RTC) plugin currently doesn't support Input Method 
 
 == TinyMCE features (plugins) support when using RTC
 
-The Real-Time Collaboration (RTC) plugin supports a subset of the features that {productname} provides.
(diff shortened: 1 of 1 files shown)


HAFRMO (approved):

modules/ROOT/pages/rtc-supported-functionality.adoc
```diff
@@ -18,6 +18,10 @@ The Real-Time Collaboration (RTC) plugin supports the latest desktop versions of
 
 Real-Time Collaboration is currently not supported on mobile devices.
 
+== IME support
+
+The Real-Time Collaboration (RTC) plugin currently doesn't support Input Method Editor (IME) as a input method.
```
TheSpyder, line 23: I think keep it in the same phrasing as the above mobile devices, rather than the below tinymce features.

```suggestion
== Input support

Real-Time Collaboration does not currently support using an Input Method Editor (IME).
```

TheSpyder (approved):

modules/ROOT/pages/rtc-supported-functionality.adoc
spocke: I just felt that it was a bit repetitive. But I changed it to the suggested text.

modules/ROOT/pages/rtc-supported-functionality.adoc
```diff
@@ -18,9 +18,13 @@ The Real-Time Collaboration (RTC) plugin supports the latest desktop versions of
...
 Real-Time Collaboration is currently not supported on mobile devices.
 
+== Input support
+
+The Real-Time Collaboration (RTC) plugin currently doesn't support Input Method Editor (IME) as a input method.
+
 == TinyMCE features (plugins) support when using RTC
 
-The Real-Time Collaboration (RTC) plugin supports a subset of the features that {productname} provides.
+Real-Time Collaboration does not currently support using an Input Method Editor (IME).
```
lnewson, line 27: Should this have been changed? This section doesn't appear to be about IME and we have it above already now. I think this would have been better left how it originally was.

TheSpyder: Seems like a mistake, perhaps Spocke copied my suggestion into the wrong place

lnewson: Ahh yeah, looks like 0b5e7ae339d41ca20d610811dc7ec1cb7ffe1c9f updated the wrong paragraph.

- [ ] Revert the change to this line and fix the correct paragraph.

spocke: Yes, I did thanks for spotting this. Had to run of to a private meeting so was a bit stressed when I fixed that change.

lnewson (approved):

lostkeys (approved):

PR #2407, opened 2022-04-26
INT-2812: Rewrite the react self-hosted docs

tiny-james:
Related Ticket: 

Description of Changes:

- Split react self-hosted docs into 4 (instead of 2)
- Rewrote and tested the whole process
- Reviewed and fixed the technical reference

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/nav.adoc
+++ b/modules/ROOT/nav.adoc
@@ -16,8 +16,10 @@
 **** xref:rails-cloud.adoc[Ruby on Rails]
 *** Self-hosted
 **** React
-***** xref:react-pm.adoc[Using a package manager]
-***** xref:react-zip.adoc[Using a .zip package]
+***** xref:react-pm-host.adoc[Using a package manager with hosting]
+***** xref:react-pm-bundle.adoc[Using a package manager with bundling]
+***** xref:react-zip-host.adoc[Using a .zip package with hosting]
+***** xref:react-zip-bundle.adoc[Using a .zip package with bundling]
 **** Angular
 ***** xref:angular-pm.adoc[Using a package manager]
 ***** xref:angular-zip.adoc[Using a .zip package]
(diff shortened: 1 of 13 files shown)


modules/ROOT/nav.adoc
```diff
@@ -16,8 +16,10 @@
 **** xref:rails-cloud.adoc[Ruby on Rails]
 *** Self-hosted
 **** React
-***** xref:react-pm.adoc[Using a package manager]
-***** xref:react-zip.adoc[Using a .zip package]
+***** xref:react-pm-host.adoc[Using a package manager with hosting]
+***** xref:react-pm-bundle.adoc[Using a package manager with bundling]
+***** xref:react-zip-host.adoc[Using a .zip package with hosting]
+***** xref:react-zip-bundle.adoc[Using a .zip package with bundling]
```
lnewson, line 22: This reads a little strange but I don't really have any suggestions. The best I can think of is swapping the hosting/bundling to be first, e.g. `Hosting using a package manager` and `Bundling using a package manager`

lnewson (approved):

modules/ROOT/partials/integrations/react-quick-start.adoc
```diff
@@ -23,22 +27,105 @@ npx create-react-app tinymce-react-demo
...
+  const log = () => {
+    if (editorRef.current) {
+      console.log(editorRef.current.getContent());
+    }
+  };
+  return (
+    <>
+      <Editor
+        apiKey='your-api-key'
+        onInit={(evt, editor) => editorRef.current = editor}
```
tiny-ben-tran, line 56:
```suggestion
        onInit={ (evt, editor) => editorRef.current = editor }
```
For better readability

modules/ROOT/nav.adoc
tiny-james: The problem is then it starts different to all the other headings that say just "Using a package manager" and "Using a .zip package".

modules/ROOT/partials/integrations/react-quick-start.adoc
tiny-james: All react code style guides I could find disagree with this spacing.

tiny-james: For example this:
https://dev.to/abrahamlawson/react-style-guide-24pp
Has this example:
```jsx
function ItemList(props) {
  return (
    <ul>
      {props.items.map((item, index) => (
        <Item
          key={item.key}
          onClick={(event) => { doSomethingWith(event, item.name, index); }}
        />
      ))}
    </ul>
  );
}
```

tiny-ben-tran (approved):
On line 633 of `react-quick-start.adoc` file, I think the markup syntax is incorrect for `+http://localhost:<port>/my_react_application+`

lnewson:
> On line 633 of `react-quick-start.adoc` file, I think the markup syntax is incorrect for `+http://localhost:<port>/my_react_application+`

This looks correct to me, as this will keep it as plain text instead of being transformed into a link.

metricjs (approved):

PR #2415, opened 2022-05-03
DOC-1642: Release notes for 6.0.2

ltrouton:
Related Ticket: DOC-1642

Description of Changes:
* Added release notes for the 6.0.2 patch release.

Pre-checks:

Review:


modules/ROOT/pages/6.0.2-release-notes.adoc
```diff
@@ -0,0 +1,65 @@
...
+* Multiple inline toolbars were shown if focused too quickly.
+* Added background and additional spacing for the text labeled buttons in the toolbar to improve visual clarity.
+* Toolbar split buttons with text used an incorrect width on touch devices.
+
+[[known-issues]]
+== Known issues
+
+This section describes issues that users of {productname} 6.0.2 may encounter and possible workarounds for these issues.
+
+There are no known issues affecting {productname} 6.0.2.
```
lnewson, line 65: This isn't accurate, as there are definitely known issues, there's just no new ones. We normally omit that section in that case, but at the same time this seems like it could be useful?

If we're going to keep this, then maybe something like this:
```suggestion
There are no new known issues affecting {productname} 6.0.2.
```

modules/ROOT/pages/release-notes.adoc
```diff
@@ -17,4 +17,12 @@ Release notes for {productname} 6.0
 
 a|
```
lnewson, line 19: This isn't rendering correctly, as you need an even amount of cells. You have 3 now and the top right is empty when it should have 6.0.2. So in this case we probably should delete these lines.

- [ ] Fix table

modules/ROOT/nav.adoc
```diff
@@ -386,5 +386,10 @@
 **** xref:6.0-release-notes-core-changes.adoc[Core changes]
 **** xref:6.0-release-notes-premium-changes.adoc[Premium changes]
 **** xref:6.0-release-notes-known-issues.adoc[Known issues]
+*** TinyMCE 6.0.2
+**** xref:6.0.2-release-notes.adoc#overview[Overview]
+**** xref:6.0.2-release-notes.adoc#general-bug-fixes[Core changes]
+**** xref:6.0.2-release-notes.adoc#accompanying-premium-plugin-changes[Premium changes]
+**** xref:6.0.2-release-notes.adoc#known-issues[Known issues]
```
lnewson, line 393: These are in the wrong place based on 5.x, as we list the versions in newest -> oldest. As such these should be above 6.0 instead of below.

- [ ] Fix nav

modules/ROOT/pages/6.0.2-release-notes.adoc
```diff
@@ -0,0 +1,65 @@
+
```
lnewson, line 1:
```suggestion
```
There shouldn't be a blank line here.

modules/ROOT/pages/6.0.2-release-notes.adoc
```diff
@@ -0,0 +1,65 @@
...
+The following premium plugin updates were released alongside {productname} 6.0.2.
+
+=== Spell Checker Pro 3.0.2
+
+The {productname} 6.0.2 release includes an accompanying release of the **Spell Checker Pro** premium plugin.
+
+**Spell Checker Pro** 3.0.2 provides the following bug fixes:
+
+* Editor focus was incorrectly changed when spellchecking multiple editors after scrolling.
```
lnewson, line 30: The old rule Tyler had was we shouldn't have a list with just one item, so maybe something like this instead (just using what I used in 5.10.4)?
```suggestion
**Spell Checker Pro** 3.0.2 fixes the page focus incorrectly changing when spellchecking is triggered on multiple editors after scrolling.
```

lnewson (approved):

modules/ROOT/pages/release-notes.adoc
lnewson: The other option is to change the `cols` so it only has a single column. However, this was setup intending to have 2 columns (like other pseudo index pages).

modules/ROOT/pages/6.0.2-release-notes.adoc
ltrouton: I mainly kept the section so we could have something to link to for the left sidebar nav. You are right that saying 'new' is more accurate so will add that. Thanks

modules/ROOT/pages/release-notes.adoc
ltrouton: Should hopefully be fixed now

modules/ROOT/nav.adoc
ltrouton: Should be fixed now. Thanks

metricjs (approved):

TheSpyder (approved):

modules/ROOT/pages/6.0.2-release-notes.adoc
```diff
@@ -0,0 +1,62 @@
...
+[[accompanying-premium-plugin-changes]]
+== Accompanying Premium Plugin changes
+
+The following premium plugin updates were released alongside {productname} 6.0.2.
+
+=== Spell Checker Pro 3.0.2
+
+The {productname} 6.0.2 release includes an accompanying release of the **Spell Checker Pro** premium plugin.
+
+**Spell Checker Pro** 3.0.2 fixes the page focus incorrectly changing when spellchecking is triggered on multiple editors after scrolling.
```
ltrouton, line 27:
```suggestion
In the previous release, **Spell Checker Pro** would change page focus incorrectly when spellchecking was triggered on multiple editors after a `scroll` event.

**Spell Checker Pro** 3.0.2, which accompanies this release, corrects this so that the page focus no longer moves in this circumstance.
```

PR #2468, opened 2022-07-12
DOC-1759: Updated end_container_on_empty_block

MitchCTiny:
Related Ticket: DOC-1759

Description of Changes:
* Updated type and default value of `end_container_on_empty_block` option

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
+++ b/modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
@@ -1,13 +1,13 @@
 [[end_container_on_empty_block]]
 == `+end_container_on_empty_block+`
 
-This option allows you to to split the current container block element if the enter key is pressed inside an empty inner block element.
+This option allows you to split the current container block element if the enter key is pressed inside an empty inner block element, and specify the type of element created.
 
-*Type:* `+Boolean+`
+*Type:* `+Boolean+` or `+String+`
 
-*Default value:* `+false+`
+*Default value:* `+'blockquote'+`
 
-*Possible values:* `+true+`, `+false+`
+*Possible values:* `+true+`, `+false+`, `+String+`
 
 === Example: Using `+end_container_on_empty_block+`
 
@@ -15,6 +15,6 @@ This option allows you to to split the current container block element if the en
 ----
 tinymce.init({
   selector: 'textarea',  // change this value according to your HTML
-  end_container_on_empty_block: true
+  end_container_on_empty_block: 'blockquote'
 });
 ----
--- a/modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
+++ b/modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
@@ -11,7 +11,7 @@ This option specifies how the current container block element is split when the 
 
 *Default value:* `+'blockquote'+`
 
-*Possible values:* `+true+`, `+false+` or comma separated `+String+` of element tag names
+*Possible values:* `+true+`, `+false+` or a comma separated `+String+` of element tag names
 
 NOTE: Prior to {productname} 6.1, this option only accepted `+Boolean+` values.
 
@@ -21,6 +21,6 @@ NOTE: Prior to {productname} 6.1, this option only accepted `+Boolean+` values.
 ----
 tinymce.init({
   selector: 'textarea',  // change this value according to your HTML
-  end_container_on_empty_block: 'blockquote'
+  end_container_on_empty_block: true
 });
 ----
--- a/modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
+++ b/modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
@@ -1,7 +1,7 @@
 [[end_container_on_empty_block]]
 == `+end_container_on_empty_block+`
 
-This option allows you to split the current container block element if the enter key is pressed inside an empty inner block element, and specify the type of element created.
+This option allows you to split the current container block element and specify the type of element created if the enter key is pressed inside an empty inner block element.
 
 *Type:* `+Boolean+` or `+String+`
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
```diff
@@ -1,20 +1,20 @@
 [[end_container_on_empty_block]]
 == `+end_container_on_empty_block+`
 
-This option allows you to to split the current container block element if the enter key is pressed inside an empty inner block element.
+This option allows you to split the current container block element if the enter key is pressed inside an empty inner block element, and specify the type of element created.
```
brian-forte, line 4:
```suggestion
This option allows you to split the current container block element and specify the type of element created if the enter key is pressed inside an empty inner block element.
```

I moved the two option-related phrases together.

The original has an unstated but easily inferred ‘then’ adverb. That is, it’s very easy to read the original as meaning

> This option allows you to split the current container block
> element if the enter key is pressed inside an empty inner block
> element, and _then_ specify the type of element created.


This reading — which makes the inferred causal relationship explicit — turns the option into something that acts as follows:

To use `end_container_on_empty_block` to split the current container

1. put the insertion point inside an empty inner block element

then

2. press *Enter*

then

3. specify the type of element to be created.

Given the option is set with a pre-chosen value, this inferred causal relationship is incorrect.

And moving the phrase makes this inference much harder to make.

tiny-ben-tran (approved):

modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
```diff
@@ -1,20 +1,20 @@
 [[end_container_on_empty_block]]
 == `+end_container_on_empty_block+`
 
-This option allows you to to split the current container block element if the enter key is pressed inside an empty inner block element.
+This option allows you to split the current container block element and specify the type of element created if the enter key is pressed inside an empty inner block element.
```
lnewson, line 4: This change is incorrect... it doesn't let you specify the type of element created. When using a string it lets you specify which blocks should be split when the enter key is pressed.

modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
```diff
@@ -1,20 +1,20 @@
...
 
-*Type:* `+Boolean+`
+*Type:* `+Boolean+` or `+String+`
 
-*Default value:* `+false+`
+*Default value:* `+'blockquote'+`
 
-*Possible values:* `+true+`, `+false+`
+*Possible values:* `+true+`, `+false+`, `+String+`
```
lnewson, line 10: This isn't correct... it's not any string, it's a comma separated list of element names.
```suggestion
*Possible values:* `+true+`, `+false+` or a comma separated `+String+` of element tag names
```

modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
lnewson: We also need to include somewhere that the string value only works in 6.1 or above.

lnewson (changes requested):

modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
```diff
@@ -1,20 +1,22 @@
 [[end_container_on_empty_block]]
 == `+end_container_on_empty_block+`
 
-This option allows you to to split the current container block element if the enter key is pressed inside an empty inner block element.
+This option allows you to split the current container block element when the enter key is pressed, and specify which blocks should be split.
```
lnewson, line 4: This change still isn't correct sorry @MitchCTiny. It's not something like allows in addition to, it's one or the other.
```suggestion
This option allows specifying how the current container block element is split when the enter key is pressed. It can either, always split the block, split for specific block elements or not split at all.
```

I think this should cover it better, but not too sure on the wording so I'd like to get @brian-forte's input here.

brian-forte, line 4: A minor edit to @lnewson’s edit plus some presentational changes:
```suggestion
This option specifies how the current container block element is split when the Enter key is pressed inside an empty inner block element. The available options are:

* always split the block;
* split for specific block elements; or
* do not split the block at all.
```
When discussing a lists of options, it’s always worth considering if presenting said options as a literal list helps.

I think it does here. The first sentence above document how and when. And the second sentence sets out what, and lays out the what in a list for clarity’s sake.

modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
```diff
@@ -1,20 +1,26 @@
...
+NOTE: Prior to {productname} 6.1, this option only accepted `+Boolean+` values.
 
 === Example: Using `+end_container_on_empty_block+`
 
 [source,js]
 ----
 tinymce.init({
   selector: 'textarea',  // change this value according to your HTML
-  end_container_on_empty_block: true
+  end_container_on_empty_block: 'blockquote'
```
lnewson, line 24: This change shouldn't have been done as we're supposed to show a different value to the default in the examples.
```suggestion
  end_container_on_empty_block: true
```

lnewson (approved):

modules/ROOT/partials/configuration/end_container_on_empty_block.adoc
```diff
@@ -1,20 +1,26 @@
...
+* split for specific block elements; or
+* do not split the block at all.
 
-*Default value:* `+false+`
+*Type:* `+Boolean+` or `+String+`
 
-*Possible values:* `+true+`, `+false+`
+*Default value:* `+'blockquote'+`
+
+*Possible values:* `+true+`, `+false+` or comma separated `+String+` of element tag names
```
lnewson, line 14: Oops sorry Mitch, just noticed I had a grammar error in the suggestion I gave previously 😞 
```suggestion
*Possible values:* `+true+`, `+false+` or a comma separated `+String+` of element tag names
```

PR #2471, opened 2022-07-12
DOC-1659: Document support for commenting on certain blocks

ltrouton:
Related Ticket: DOC-1659

Description of Changes:
* Generalise specific uses of the word 'text' to 'content'
* Add section for supported content (unsure if this is the best location)

Pre-checks:

Review:


modules/ROOT/pages/introduction-to-tiny-comments.adoc
```diff
@@ -37,6 +37,20 @@ The Comments plugin allows the user to perform the following functions:
 * Resolve a comment thread
 * Delete a comment or comment thread
 
+=== Supported content
```
lnewson, line 40: Does this need to be in the `annotations.adoc` page as well? If so we should make this a partial and include it in both places.

modules/ROOT/pages/introduction-to-tiny-comments.adoc
```diff
@@ -37,6 +37,20 @@ The Comments plugin allows the user to perform the following functions:
...
+In addition, the user can comment on various blocks, including:
+
+NOTE: This feature is only available for {productname} 6.1 and later.
+
+* Tables of contents created using the xref:tableofcontents.adoc[Table of Contents] plugin.
+* Images and images with captions created using the xref:image.adoc[Image] plugin.
+* Code samples created using the xref:codesample.adoc[Code Sample] plugin.
+* Media created using the xref:media.adoc[Media] plugin and xref:introduction-to-mediaembed.adoc[Enhanced Media Embed] plugin.
+* Embedded pages created using the xref:pageembed.adoc[Page Embed] plugin.
+
```
MitchCTiny, line 53: Should the note go below the content types, so as not to disrupt the flow of the content?

```suggestion
* Tables of contents created using the xref:tableofcontents.adoc[Table of Contents] plugin.
* Images and images with captions created using the xref:image.adoc[Image] plugin.
* Code samples created using the xref:codesample.adoc[Code Sample] plugin.
* Media created using the xref:media.adoc[Media] plugin and xref:introduction-to-mediaembed.adoc[Enhanced Media Embed] plugin.
* Embedded pages created using the xref:pageembed.adoc[Page Embed] plugin.

NOTE: This feature is only available for {productname} 6.1 and later.

```

modules/ROOT/pages/introduction-to-tiny-comments.adoc
```diff
@@ -37,6 +37,20 @@ The Comments plugin allows the user to perform the following functions:
 * Resolve a comment thread
 * Delete a comment or comment thread
 
+=== Supported content
+
+The Comments plugin allows the user to comment on text.
+
+In addition, the user can comment on various blocks, including:
+
+NOTE: This feature is only available for {productname} 6.1 and later.
```
lnewson, line 46: This shouldn't be done like this... we should create a partial and include it instead. See https://github.com/tinymce/tinymce-docs/blob/release/TinyMCE_5.10/modules/ROOT/partials/misc/requires_5_9v.adoc for an example. Additionally, this is quite odd being between `In addition, the user can comment on various blocks, including:` which is the start of the list and the actual list.

- [ ] Fix admonition warning
- [ ] Move above start of list instead of in the middle

MitchCTiny (approved):

lnewson (approved):

modules/ROOT/pages/introduction-to-tiny-comments.adoc
ltrouton: I have split out the supported content into its own file and added it to the annotation and comments pages.

modules/ROOT/pages/introduction-to-tiny-comments.adoc
ltrouton: Yeah, that would be better. Thanks

modules/ROOT/pages/introduction-to-tiny-comments.adoc
ltrouton: Admonition partial has been added. I also reworked the structure of the supported content section so it now looks like this:
image

lnewson: Hmm... I think we need that admonition above the list if we're using the above.

ltrouton: Should be better now. Thanks for that

brian-forte (approved):

PR #2467, opened 2022-07-12
DOC-1676: Added iframe_template_callback option to media

MitchCTiny:
Related Ticket: DOC-1676

Description of Changes:
* Added iframe_template_callback option to media

Pre-checks:
- [x] Branch prefixed with `feature/6/` or `hotfix/6/`
- ~[x] `modules/ROOT/nav.adoc` has been updated (if applicable)~
- [x] Files has been included where required (if applicable)
- ~[x] Files removed have been deleted, not just excluded from the build (if applicable)~
- ~[x] (New product features only) Release Note added~

Review:
- [x] Documentation Team Lead has reviewed
- [x] Product Manager has reviewed

PR #2472, opened 2022-07-12
DOC-1660: Add tinycomments.getEventLog API

MitchCTiny:
Related Ticket: DOC-1660

Description of Changes:
* Added tinycomments.getEventLog() API documentation to comments-commands-events-apis.adoc

Pre-checks:
~- [x] Files removed have been deleted, not just excluded from the build (if applicable)~
~- [x] (New product features only) Release Note added~

Review:


lnewson (changes requested):
You're missing the `CommentChanged` event here... you also need to change the title of this page and of the nav to be more than just `Commands`.

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,29 @@
...
+
+* A new comment is added (and who added it).
+* A comment is edited (and who edited it).
+* A reply to a comment is added (and who added the reply).
+* A comment is resolved (and who resolved the comment).
+* A comment is deleted (and who deleted the comment).
+
+The event log can be retrieved either in full or with the `after` option, which restricts the returned list to Comment events after the time-stamp date, both shown in the following:
+
+NOTE: This feature is only available for {productname} 6.1 and later.
```
brian-forte, line 26:
```suggestion
NOTE: This feature is only available for {productname} 6.1 and later.

The event log can be retrieved either in full or with the `after` option, which restricts the returned list to Comment events after the time-stamp date, both shown in the following:
```
I think reversing the paragraph order here puts the admonition in a less disruptive spot.

The admonition is now immediately after the feature listing (which is what requires 6.1 or later) and the ‘here’s the sample code’ sentence is now immediately adjacent the sample code.

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,30 @@
...
+
+=== Example: Using the `+comments+` plugin APIs
+
+[source,js]
+----
+const comments = tinymce.activeEditor.plugins.tinycomments;
+
+console.log(comments.getEventLog());
+console.log(comments.getEventLog(
+    { after: '<iso-8601-format-time-stamp>' }
```
TheSpyder, line 37: I feel like we need an example timestamp here, or a better explanation of what the format is.

TheSpyder (approved):

modules/ROOT/pages/comments-commands-events-apis.adoc
MitchCTiny: I'll investigate, I pulled this from the release notes

TheSpyder: ah. there's so many comments on that I've lost track.

modules/ROOT/partials/misc/minimum-version.adoc
```diff
@@ -0,0 +1 @@
+NOTE: This feature is only available for {productname} {minimumversion} and later.
```
lnewson, line 1: This isn't how this has been done in the past and having to specify the `minimumversion` all the time is messy... see https://github.com/tinymce/tinymce-docs/pull/2471#discussion_r918548089 on how this should be done for this specific case.

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,30 @@
 The Comments plugin provides the following {productname} commands.
 
 include::partial$commands/comments-cmds.adoc[]
+
+== API
+
+The Comments plugin exposes an API, `editor.plugins.tinycomments.getEventLog()`, which returns a log of all `CommentChange` events which are fired when:
```
lnewson, line 16: This also doesn't follow the standard template for this... take a look at https://www.tiny.cloud/docs/tinymce/6/a11ychecker/#apis. We should have a heading for each API so it shows in the nav at the very least.

modules/ROOT/pages/comments-commands-events-apis.adoc
brian-forte: [ISO-8601](https://en.wikipedia.org/wiki/ISO_8601) is widely used and it’s very common as the format APIs take. So I guess I was assuming just the ISO number would ring bells for a fair proportion of the likely readership (it does for me, FWIW).

An alternative would be to put a generic version of the longest 8601 standard string as placeholder text:

`<yyyy-mm-ddThh:mm:ss+00:00>`

I’m not sure that’s more readable.

If we go with an example timestamp we should add a post-example note about reading the time-stamp and about how needs to be an ISO-8601 format string.

Something like:

```
[source,js]
----
console.log(comments.getEventLog());
console.log(comments.getEventLog(
  { after: '<2022-07-12T00:12:00+00:00>' }
)
----
```

The example above restricts the returned list to Comment events created after 12:00 UTC on 2022-07-12.

The entered time-stamp must be in ISO-8601 format.

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,30 @@
...
+
+=== Example: Using the `+comments+` plugin APIs
+
+[source,js]
+----
+const comments = tinymce.activeEditor.plugins.tinycomments;
+
+console.log(comments.getEventLog());
+console.log(comments.getEventLog(
+    { after: '2022-02-22' }     //ISO-8601 standard: YYYY-MM-DD
```
TheSpyder, line 37: I suggest including some minutes, at least, since it's unlikely customers will want to use date granularity for the API calls

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,34 @@
...
+
+==== Example: Using `+getEventLog()+`
+
+[source,js]
+----
+const comments = tinymce.activeEditor.plugins.tinycomments;
+
+console.log(comments.getEventLog());
+console.log(comments.getEventLog(
+    { after: '2022-02-22T12:34:56Z' }     //ISO-8601 standard: YYYY-MM-DDThh:mm:ssZ
```
lnewson, line 41:
```suggestion
    { after: '2022-02-22T12:34:56Z' }  // ISO-8601 standard: YYYY-MM-DDThh:mm:ssZ
```

modules/ROOT/partials/misc/minimum-version.adoc
TheSpyder: We have ended up with every PR doing it separately. This was not well coordinated.

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,34 @@
...
 include::partial$commands/comments-cmds.adoc[]
+
+== APIs
+
+The {pluginname} plugin provides the following API.
+
+[[getEventLog]]
+=== getEventLog()
+
+Returns a log of all `CommentChange` events. These events are dispatched whenever:
```
lnewson, line 21: This isn't correct... the `CommentChange` is an event fired when something changes, it's not directly related to what's returned by this API. The `CommandChange` event should instead be in an `Events` heading on this page above `Commands` like this: https://www.tiny.cloud/docs/tinymce/6/introduction-to-tiny-spellchecker/#events. It also needs adding to this page: https://www.tiny.cloud/docs/tinymce/6/events/#plugin-events

modules/ROOT/partials/misc/minimum-version.adoc
MitchCTiny: I have added a new partial file in misc named requires_6_1v.adoc in this PR and https://github.com/tinymce/tinymce-docs/pull/2467, just as in the linked example.

lnewson: @MitchCTiny that's not what I said to do... I said it needs coordination, not duplication.

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,60 @@
...
+== APIs
+
+The {pluginname} plugin provides the following API.
+
+[[getEventLog]]
+=== getEventLog()
+
+include::partial$misc/admon-requires-6.1v.adoc[]
+
+Returns a log of all `CommentChange` events.
```
lnewson, line 55: @MitchCTiny as I said in my previous comment this isn't correct... it's not a log of `CommentChange` events, the two aren't the same thing. A `CommentChange` just means something has changed, it doesn't have the information. The `getEventLog()` API returns that explicit information about the comment state for the change.

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,60 @@
...
+
+=== CommentChange event
+
+This event triggers whenever:
+
+* A new comment is added (and who added it).
+* A comment is edited (and who edited it).
+* A reply to a comment is added (and who added the reply).
+* A comment is resolved (and who resolved the comment).
+* A comment is deleted (and who deleted the comment).
```
lnewson, line 26: This probably should have stay where it was in the `getEventLog` section.

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -1,6 +1,6 @@
-= Commands for the comments plugin
-:navtitle: Commands
-:description: Information on the commands provided with the comments plugin.
+= Commands and APIs for the comments plugin
```
lnewson, line 1: This needs updating again to include `Events`. It also needs updating in the `nav.adoc` file here: https://github.com/tinymce/tinymce-docs/blame/staging/docs-6/modules/ROOT/nav.adoc#L289

modules/ROOT/pages/comments-commands-events-apis.adoc
```diff
@@ -10,3 +10,60 @@
...
+
+==== Example: Using `+getEventLog()+`
+
+[source,js]
+----
+const comments = tinymce.activeEditor.plugins.tinycomments;
+
+console.log(comments.getEventLog());
+console.log(comments.getEventLog(
+    { after: '2022-02-22T12:34:56Z' }  //ISO-8601 standard: YYYY-MM-DDThh:mm:ssZ
```
lnewson, line 67: This doesn't follow our syntax style, it should be 2 spaces and there should be a space after the inline comment syntax like I had in my previous suggestion.
```suggestion
  { after: '2022-02-22T12:34:56Z' }  // ISO-8601 standard: YYYY-MM-DDThh:mm:ssZ
```

lnewson (approved):

brian-forte (approved):

PR #2492, opened 2022-08-24
DOC-1779: edited .js examples in customsidebar.adoc to ‘improve accuracy’.

brian-forte:
Edited javascript examples: onSetup now returns an empty function.

NB: `console.log` is still used despite it ‘not [being] a valid api function.’

Related Ticket: 

Description of Changes:
* edited .js examples in customsidebar.adoc to ‘improve accuracy’.

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/pages/customsidebar.adoc
+++ b/modules/ROOT/pages/customsidebar.adoc
@@ -77,11 +77,11 @@ tinymce.init({
       tooltip: 'My sidebar',
       icon: 'comment',
       onSetup: (api) => {
-      console.log('Render panel', api.element());
-      return () => {
-        console.log('Removing sidebar');
+        console.log('Render panel', api.element());
+        return () => {
+          console.log('Removing sidebar');
+        };
       };
-    };
       onShow: (api) => {
         console.log('Show panel', api.element());
         api.element().innerHTML = 'Hello world!';
--- a/modules/ROOT/pages/customsidebar.adoc
+++ b/modules/ROOT/pages/customsidebar.adoc
@@ -105,10 +105,11 @@ tinymce.PluginManager.add('myplugin', (editor) => {
     icon: 'comment',
     onSetup: (api) => {
      console.log('Render panel', api.element());
-      return () => {
-	   api.element().focus();
-	  };
-	},
+     api.element().focus();
+     return () => {
+       editor.focus();
+     };
+   },
     onShow: (api) => {
       console.log('Show panel', api.element());
       api.element().innerHTML = 'Hello world!';
--- a/modules/ROOT/pages/customsidebar.adoc
+++ b/modules/ROOT/pages/customsidebar.adoc
@@ -81,7 +81,7 @@ tinymce.init({
         return () => {
           console.log('Removing sidebar');
         };
-      };
+      },
       onShow: (api) => {
         console.log('Show panel', api.element());
         api.element().innerHTML = 'Hello world!';
--- a/modules/ROOT/pages/customsidebar.adoc
+++ b/modules/ROOT/pages/customsidebar.adoc
@@ -104,12 +104,12 @@ tinymce.PluginManager.add('myplugin', (editor) => {
     tooltip: 'My sidebar',
     icon: 'comment',
     onSetup: (api) => {
-     console.log('Render panel', api.element());
-     api.element().focus();
-     return () => {
-       editor.focus();
-     };
-   },
+      console.log('Render panel', api.element());
+      api.element().focus();
+      return () => {
+        editor.focus();
+      };
+    },
     onShow: (api) => {
       console.log('Show panel', api.element());
       api.element().innerHTML = 'Hello world!';
--- a/modules/ROOT/pages/customsidebar.adoc
+++ b/modules/ROOT/pages/customsidebar.adoc
@@ -78,9 +78,10 @@ tinymce.init({
       icon: 'comment',
       onSetup: (api) => {
         console.log('Render panel', api.element());
-			return () => {
-		 	api.element().focus();
-			};
+	api.element().focus();
+	return () => {
+	  editor.focus();
+	};
       },
       onShow: (api) => {
         console.log('Show panel', api.element());
--- a/modules/ROOT/pages/customsidebar.adoc
+++ b/modules/ROOT/pages/customsidebar.adoc
@@ -79,7 +79,7 @@ tinymce.init({
       onSetup: (api) => {
         console.log('Render panel', api.element());
 			return () => {
-		 	api.element().removeChild(box);
+		 	api.element().focus();
 			};
       },
       onShow: (api) => {
@@ -105,7 +105,7 @@ tinymce.PluginManager.add('myplugin', (editor) => {
     onSetup: (api) => {
      console.log('Render panel', api.element());
       return () => {
-	   api.element().removeChild(box);
+	   api.element().focus();
 	  };
 	},
     onShow: (api) => {
--- a/modules/ROOT/pages/customsidebar.adoc
+++ b/modules/ROOT/pages/customsidebar.adoc
@@ -102,7 +102,7 @@ tinymce.PluginManager.add('myplugin', (editor) => {
   editor.ui.registry.addSidebar('mysidebar', {
     tooltip: 'My sidebar',
     icon: 'comment',
-      onSetup: (api) => {
+    onSetup: (api) => {
       console.log('Render panel', api.element());
       return () => {
         console.log('Removing sidebar');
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/customsidebar.adoc
```diff
@@ -78,6 +78,9 @@ tinymce.init({
       icon: 'comment',
       onSetup: (api) => {
         console.log('Render panel', api.element());
+			return () => {
+		 	api.element().focus();
+			};
```
lnewson, line 83:
```suggestion
        console.log('Render panel', api.element());
        return () => api.element().focus();
```
Code examples should use spaces instead of tabs as such the indentation wasn't great here.

On that note though... why are we focusing the sidebar element when the sidebar is destroy? That doesn't make any sense.

modules/ROOT/pages/customsidebar.adoc
```diff
@@ -100,8 +103,11 @@ tinymce.PluginManager.add('myplugin', (editor) => {
     tooltip: 'My sidebar',
     icon: 'comment',
     onSetup: (api) => {
-      console.log('Render panel', api.element());
-    },
+     console.log('Render panel', api.element());
+      return () => {
+	   api.element().focus();
+	  };
+	},
```
lnewson, line 110:
```suggestion
      console.log('Render panel', api.element());
      return () => api.element().focus();
    },
```
Just fixing indentation again. Likewise though focusing the element when it's destroyed doesn't make sense to me unless I'm missing something?

modules/ROOT/pages/customsidebar.adoc
```diff
@@ -78,6 +78,9 @@ tinymce.init({
       icon: 'comment',
       onSetup: (api) => {
         console.log('Render panel', api.element());
+			return () => {
+		 	api.element().focus();
+			};
       },
```
MitchCTiny, line 84:
```suggestion
      onSetup: (api) => {
        console.log('Render panel', api.element());
	api.element().focus();
	return () => {
	  editor.focus();
	};
      },
```
The element would not be able to be focused when it is de-rendered, so something like this makes more sense. And focus on the element would make more sense as it is rendered.

As per the `Custom Group Toolbar` documentation:
`onSetup is a complex property. It takes a function that is passed the component’s API and should return a callback that is passed the component’s API and returns nothing. This occurs because onSetup runs whenever the component is rendered, and the returned callback is executed when the component is destroyed. This is essentially an onTeardown handler, and can be used to unbind events and callbacks.`

modules/ROOT/pages/customsidebar.adoc
MitchCTiny, line 110:
```suggestion
     console.log('Render panel', api.element());
     api.element().focus();
     return () => {
       editor.focus();
     };
   },
```

modules/ROOT/pages/customsidebar.adoc
brian-forte: I’ve committed @MitchCTiny’s suggested edits. They fix the indent character errors and change the sample to correct the focus error.

modules/ROOT/pages/customsidebar.adoc
brian-forte: As per above, I’ve committed @MitchCTiny’s suggested edits. They fix the indent character errors and change the sample to correct the focus error.

modules/ROOT/pages/customsidebar.adoc
```diff
@@ -100,8 +104,12 @@ tinymce.PluginManager.add('myplugin', (editor) => {
...
     icon: 'comment',
     onSetup: (api) => {
-      console.log('Render panel', api.element());
-    },
+     console.log('Render panel', api.element());
+     api.element().focus();
+     return () => {
+       editor.focus();
+     };
+   },
```
MitchCTiny, line 112:
```suggestion
      console.log('Render panel', api.element());
      api.element().focus();
      return () => {
        editor.focus();
      };
    },
```
Sorry Brian, I must have messed up the indentation by a single space, this should fix it.

MitchCTiny (approved):

modules/ROOT/pages/customsidebar.adoc
```diff
@@ -78,6 +78,10 @@ tinymce.init({
       icon: 'comment',
       onSetup: (api) => {
         console.log('Render panel', api.element());
+	api.element().focus();
```
tiny-ben-tran, line 81: What is the purpose of `api.element().focus()` since the code puts the focus back on the editor on return?

tiny-ben-tran (approved):

modules/ROOT/pages/customsidebar.adoc
tiny-ben-tran: I assume that the `api.element().focus()` is to put focus on the sidebar when it is shown on `init`

ephox-mogran (commented):
I'm not sure about calling focus inside `onSetup`. The main thing that has to be fixed was that onSetup has to return a function. The reason for this is that it should return its own teardown/destroy handler.

So I'd recommend getting rid of the `api.element().focus` calls because you probably don't always want the sidebar to grab focus, and it may not be focusable itself. Instead, just make sure it returns a function.

So something like: 

```
onSetup: (api) => {
      console.log('Render panel', api.element());
      return () => {
        console.log('Removing sidebar');
      };
    }
```

In general, I'd advise against adding focus calls to these demos, because focus is a nuanced topic and greatly depends on the UX desired. And these demos will just be copied and pasted, so we want them to be as harmless as possible. That's my view, anyway.

brian-forte:
@ephox-mogran:

To quote my commit comment:

> Sample javascripts no longer calls focus, as per [the argument in this comment in the pull request](https://github.com/tinymce/tinymce-docs/pull/2492#pullrequestreview-1091310946).

lnewson:
@brian-forte unfortunately your latest change has broken all the indentation again... I'll log a suggestion but you might need to sort out what's going on with your editor.

modules/ROOT/pages/customsidebar.adoc
```diff
@@ -77,8 +77,11 @@ tinymce.init({
       tooltip: 'My sidebar',
       icon: 'comment',
       onSetup: (api) => {
-        console.log('Render panel', api.element());
-      },
+      console.log('Render panel', api.element());
+      return () => {
+        console.log('Removing sidebar');
+      };
+    };
```
lnewson, line 84:
```suggestion
        console.log('Render panel', api.element());
        return () => {
          console.log('Removing sidebar');
        };
      };
```

modules/ROOT/pages/customsidebar.adoc
```diff
@@ -99,8 +102,11 @@ tinymce.PluginManager.add('myplugin', (editor) => {
   editor.ui.registry.addSidebar('mysidebar', {
     tooltip: 'My sidebar',
     icon: 'comment',
-    onSetup: (api) => {
+      onSetup: (api) => {
```
lnewson, line 105:
```suggestion
    onSetup: (api) => {
```

modules/ROOT/pages/customsidebar.adoc
```diff
@@ -78,7 +78,10 @@ tinymce.init({
       icon: 'comment',
       onSetup: (api) => {
         console.log('Render panel', api.element());
-      },
+        return () => {
+          console.log('Removing sidebar');
+        };
+      };
```
lnewson, line 84:
```suggestion
      },
```
Oops somehow this has been converted into a semicolon when it should be a comma.

brian-forte: That could be a typo on my part. Commiting.

lnewson (approved):

brian-forte:
> you might need to sort out what's going on with your editor.

Just spent some quality time in BBEdit’s preferences.

Hitting `tab` (which is my habit of decades standing when it comes to indenting a line, so that’s not going away anytime soon) will now _auto-expand_ to **one** space (not four, which is BBEdit’s default).

It feels wrong, but I think this is, now, a tiny-friendly setting.

lnewson:
> > you might need to sort out what's going on with your editor.
> 
> Just spent some quality time in BBEdit’s preferences.
> 
> Hitting `tab` (which is my habit of decades standing when it comes to indenting a line, so that’s not going away anytime soon) will now _auto-expand_ to **one** space (not four, which is BBEdit’s default).
> 
> It feels wrong, but I think this is, now, a tiny-friendly setting.

Setting it to 2 spaces would probably be better as that's our normal indentation style for JS, etc... It might feel less "wrong" but if you're used to 4 I somewhat doubt it 😞

PR #2514, opened 2022-09-21
DOC-1768: Document `format_noneditable_selector` option

ltrouton:
Related Ticket: DOC-1768, TINY-8905

Description of Changes:
* Document new `format_noneditable_selector` option added as part of TinyMCE 6.2

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
+++ b/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
@@ -1,0 +1,20 @@
+[[format_noneditable_selector]]
+== `+format_noneditable_selector+`
+
+This option allows a non-editable (`+contenteditable="false"+`) element that matches the selector specified to be wrapped with a valid format.
+
+NOTE: The non-editable element must not have any editable descendants in order for it to be wrappable.
+
+*Type:* `+String+`
+
+*Default value:* `+''+`
+
+=== Example: Using `+format_noneditable_selector+`
+
+[source,js]
+----
+tinymce.init({
+  selector: 'textarea',  // change this value according to your HTML
+  format_noneditable_selector: 'span,p.special'
+});
+----
--- a/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
+++ b/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
@@ -3,11 +3,11 @@
 
 This option allows a non-editable (`+contenteditable="false"+`) element that matches the selector specified to be wrapped with a format.
 
-Formats that can wrap include block formats that have the `+wrapper+` parameter set to `+true+` and inline formats. Selector formats are not supported.
-
 In addition to a non-editable element matching the selector, any non-editable element that has the `+data-mce-cef-wrappable+` attribute can be wrapped with a format.
 
-NOTE: The non-editable element must not have any editable descendants in order for it to be wrappable.
+Formats that can wrap include block formats that have the `+wrapper+` parameter set to `+true+` and inline formats. Selector formats are not supported.
+
+NOTE: The non-editable element must not have any editable (`+contenteditable="true"+`) descendants in order for it to be wrappable.
 
 *Type:* `+String+`
 
--- a/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
+++ b/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
@@ -1,13 +1,15 @@
 [[format_noneditable_selector]]
 == `+format_noneditable_selector+`
 
-This option allows a non-editable (`+contenteditable="false"+`) element that matches the selector specified to be wrapped with a format.
+If an element has a `contenteditable="false"` attribute, the `format_noneditable_selector` allows that element to be wrapped with a format if the element matches the specified selector.
 
-In addition to a non-editable element matching the selector, any non-editable element that has the `+data-mce-cef-wrappable+` attribute can be wrapped with a format.
+As well, an element with a `contenteditable="false"` attribute that also has the `data-mce-cef-wrappable` attribute can be wrapped with a format.
 
-Formats that can wrap include block formats that have the `+wrapper+` parameter set to `+true+` and inline formats. Selector formats are not supported.
+Wrappable formats include block formats that have the `wrapper` parameter set to `true` and inline formats.
 
-NOTE: The non-editable element must not have any editable (`+contenteditable="true"+`) descendants in order for it to be wrappable.
+Selector formats are not supported.
+
+NOTE: elements with `contenteditable="false"` attributes that contain descendant elements with `contenteditable="true"` attributes are not wrappable.
 
 *Type:* `+String+`
 
--- a/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
+++ b/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
@@ -1,13 +1,15 @@
 [[format_noneditable_selector]]
 == `+format_noneditable_selector+`
 
-This option allows a non-editable (`+contenteditable="false"+`) element that matches the selector specified to be wrapped with a valid format.
+This option allows a non-editable (`+contenteditable="false"+`) element that matches the selector specified to be wrapped with a format.
+
+Formats that can wrap include block formats that have the `+wrapper+` parameter set to `+true+` and inline formats. Selector formats are not supported.
+
+In addition to a non-editable element matching the selector, any non-editable element that has the `+data-mce-cef-wrappable+` attribute can be wrapped with a format.
 
 NOTE: The non-editable element must not have any editable descendants in order for it to be wrappable.
 
 *Type:* `+String+`
-
-*Default value:* `+''+`
 
 === Example: Using `+format_noneditable_selector+`
 
--- a/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
+++ b/modules/ROOT/partials/configuration/format_noneditable_selector.adoc
@@ -1,7 +1,7 @@
 [[format_noneditable_selector]]
 == `+format_noneditable_selector+`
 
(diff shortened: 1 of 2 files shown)


modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,20 @@
+[[format_noneditable_selector]]
+== `+format_noneditable_selector+`
+
+This option allows a non-editable (`+contenteditable="false"+`) element that matches the selector specified to be wrapped with a valid format.
```
lnewson, line 4: This doesn't allow all formats from memory. It only allows inline elements to work IIRC? As you can't use a selector format for example as it's not wrappable.

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,20 @@
...
+
+NOTE: The non-editable element must not have any editable descendants in order for it to be wrappable.
+
+*Type:* `+String+`
+
+*Default value:* `+''+`
```
lnewson, line 10:
```suggestion
```

Don't normally document nullable/blank values since they offer no purpose 🤷

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,20 @@
+[[format_noneditable_selector]]
+== `+format_noneditable_selector+`
```
lnewson, line 2: Do you need to mention anything about `data-mce-cef-wrappable` here? As that seems important to know it's actually part of the defaults that cannot be changed and applies automatically to elements with that attribute.

ltrouton: Yeah, I wasn't sure whether to mention it or not. I guess since it can't be overridden in any way by the integrator, then having a note about it would be good.

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
ltrouton: Technically it can apply to non-editable blocks if the format is block type with `wrapper: true`. The two defaults would be applying blockquote and div. Then it can wrap an inline non-editable element with any inline format assuming it is valid as per the schema.

I'll see if I can put something together without being too wordy.

ltrouton: Does adding something like this seem suitable?
```
Valid formats are block formats that have the `+wrapper+` parameter set to `+true+` and inline formats. These can be applied to non-editable block elements such as `+<p>+` and non-editable inline elements such as `+<span>+` respectively.

Selector formats are not supported.
```

lnewson: I don't think mentioning anything about `valid formats` is a good idea, as valid formats just means it's a format that TinyMCE accepts. What we really want here is `formats that allow wrapping` or similar.

ltrouton: Good point, that would be a better way of phrasing it.

ltrouton: I have pushed up some changes. Hopefully, it is closer to want we should say.

lnewson: That seems reasonable to me.

MitchCTiny (approved):

tiny-ben-tran (approved):

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+This option allows a non-editable (`+contenteditable="false"+`) element that matches the selector specified to be wrapped with a format.
+
+Formats that can wrap include block formats that have the `+wrapper+` parameter set to `+true+` and inline formats. Selector formats are not supported.
+
+In addition to a non-editable element matching the selector, any non-editable element that has the `+data-mce-cef-wrappable+` attribute can be wrapped with a format.
```
ephox-mogran, line 8: The `format` mentioned here for the `data-mce-cef-wrappable` ... is that limited by the statement above, or can *all* formats wrap non-editable elements with `data-mce-cef-wrappable`? I assume it's still limited, but I'm not sure if that's clear.

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+NOTE: The non-editable element must not have any editable descendants in order for it to be wrappable.
```
ephox-mogran, line 10: Given that we define what a `non-editable` is (i.e. `contenteditable=false`), do we need to define what an `editable` is, or should it be self-explanatory?

ephox-mogran (approved):

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
spocke: Interesting thought this was internal. From what I can see there is no mention of this new attribute in the changelog. It only mentions the selector. We have not documented any other internal attributes from what I know. But this one doesn't seem to be stripped by default on save so I guess this is a special case. So is the intention that we want this to be public?

- [ ] Confirm that this attribute is to be public knowledge

spocke (approved):

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,22 @@
...
+
+*Type:* `+String+`
+
+=== Example: Using `+format_noneditable_selector+`
+
+[source,js]
+----
+tinymce.init({
+  selector: 'textarea',  // change this value according to your HTML
+  format_noneditable_selector: 'span,p.special'
```
lorenzo-pomili, line 20:
```suggestion
  format_noneditable_selector: 'span, p.special'
```
should there be space here?

ltrouton: Normally we don't put a space when defining a selector so would say it isn't needed

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
ltrouton: Probably can't hurt to do it that as well

ltrouton: Have added it

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
ltrouton: I have flipped the paragraphs around which hopefully makes it clearer.

With regards to whether to mention `data-mce-cef-wrappable` or not. I think the fact that like you say it is not ephemeral and also is something that is always enforced i.e not integrator configurable, it is probably good for people to be aware of it in some capacity.

It probably should have been mentioned in the changelog but wasn't.

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,22 @@
+[[format_noneditable_selector]]
+== `+format_noneditable_selector+`
+
+This option allows a non-editable (`+contenteditable="false"+`) element that matches the selector specified to be wrapped with a format.
+
+In addition to a non-editable element matching the selector, any non-editable element that has the `+data-mce-cef-wrappable+` attribute can be wrapped with a format.
+
+Formats that can wrap include block formats that have the `+wrapper+` parameter set to `+true+` and inline formats. Selector formats are not supported.
+
+NOTE: The non-editable element must not have any editable (`+contenteditable="true"+`) descendants in order for it to be wrappable.
```
brian-forte, line 10:
```suggestion
If an element has a `contenteditable="false"` attribute, the `format_noneditable_selector` allows that element to be wrapped with a format if the element matches the specified selector.

As well, an element with a `contenteditable="false"` attribute that also has the `data-mce-cef-wrappable` attribute can be wrapped with a format.

Wrappable formats include block formats that have the `wrapper` parameter set to `true` and inline formats.

Selector formats are not supported.

NOTE: elements with `contenteditable="false"` attributes that contain descendant elements with `contenteditable="true"` attributes are not wrappable.
```
The above edit makes it explicit `contenteditable` is an attribute an element can take.

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,24 @@
...
+
+If an element has a `contenteditable="false"` attribute, the `format_noneditable_selector` allows that element to be wrapped with a format if the element matches the specified selector.
+
+As well, an element with a `contenteditable="false"` attribute that also has the `data-mce-cef-wrappable` attribute can be wrapped with a format.
+
+Wrappable formats include block formats that have the `wrapper` parameter set to `true` and inline formats.
+
+Selector formats are not supported.
+
+NOTE: elements with `contenteditable="false"` attributes that contain descendant elements with `contenteditable="true"` attributes are not wrappable.
```
ltrouton, line 12:
```suggestion
NOTE: Elements with `contenteditable="false"` attributes that contain descendant elements with `contenteditable="true"` attributes are not wrappable.
```

modules/ROOT/partials/configuration/format_noneditable_selector.adoc
```diff
@@ -0,0 +1,24 @@
+[[format_noneditable_selector]]
+== `+format_noneditable_selector+`
+
+If an element has a `contenteditable="false"` attribute, the `format_noneditable_selector` allows that element to be wrapped with a format if the element matches the specified selector.
```
ltrouton, line 4:
```suggestion
If an element has a `contenteditable="false"` attribute, the `format_noneditable_selector` option allows that element to be wrapped with a format if the element matches the specified selector.
```

PR #2518, opened 2022-09-21
DOC-1774: Update autocompleter `trigger` option

MitchCTiny:
Related Ticket: DOC-1774

Description of Changes:
* Update name of `ch` single character trigger option to `trigger` option with multiple character (string) support

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/pages/autocompleter.adoc
+++ b/modules/ROOT/pages/autocompleter.adoc
@@ -23,12 +23,12 @@ The two arguments this method take are:
 [cols="1,2,1,4",options="header"]
 |===
 |Name |Value |Requirement |Description
-|ch |`+string+` |Required |The character to trigger the autocompleter.
+|ch |`+string+` (of one character) |Required |The character to trigger the autocompleter.
 include::partial$misc/admon-deprecated-6.2v.adoc[]
-It has been replaced with the `+trigger+` option.
+It has been replaced with the `+trigger+` option, which should be used instead.
 |trigger |`+string+` |Required |The string to trigger the autocompleter.
 include::partial$misc/admon-requires-6.2v.adoc[]
-It replaces the deprecated `+ch+` option.
+It replaces the deprecated `+ch+` option, and should be used instead of that option.
 |fetch |`+(pattern: string, maxResults: number, fetchOptions: Record<string, any>) => Promise<AutocompleteItem[] \| CardMenuItem[]+` |Required |A function that is passed the current matched text pattern, the maximum number of expected results and any additional fetch options. The function should return a Promise containing matching results.
 |onAction |`+(api, rng: Range, value: string) => void+` |Required |A function invoked when a fetched item is selected.
 |columns |number or `'auto'` |Optional |default: auto - The number of columns to show. If set to `+1+` column, then icons and text are displayed, otherwise only icons are displayed.
--- a/modules/ROOT/pages/autocompleter.adoc
+++ b/modules/ROOT/pages/autocompleter.adoc
@@ -28,7 +28,6 @@ include::partial$misc/admon-deprecated-6.2v.adoc[]
 It has been replaced with the `+trigger+` option, which should be used instead.
 |trigger |`+string+` |Required |The string to trigger the autocompleter.
 include::partial$misc/admon-requires-6.2v.adoc[]
-It replaces the deprecated `+ch+` option, and should be used instead of that option.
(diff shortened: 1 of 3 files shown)


tiny-ben-tran (approved):

lnewson (dismissed):
You can't just remove the option. You need to mention it's deprecated in the relevant version and the replacement. You also need to mention the new flag only works in newer versions as remember this is the docs for all of 6.x, not just 6.2.

MitchCTiny:
> You can't just remove the option. You need to mention it's deprecated in the relevant version and the replacement. You also need to mention the new flag only works in newer versions as remember this is the docs for all of 6.x, not just 6.2.

Should I also mention that it is slated for remnov

modules/ROOT/pages/autocompleter.adoc
```diff
@@ -23,7 +23,12 @@ The two arguments this method take are:
 [cols="1,2,1,4",options="header"]
 |===
 |Name |Value |Requirement |Description
-|ch |string (of one character) |Required |The character to trigger the autocompleter.
+|ch |`+string+` |Required |The character to trigger the autocompleter.
```
lnewson, line 26: You haven't fixed this fully... it's still missing the part about trigger replaced this and should be used instead.

- [ ] Correct details

lnewson: I see actually you have this in `trigger` actually... it'd make more sense to have it here, as this is the option we don't want people to use moving forward.

MitchCTiny: It is here, this table cell follows three lines as the partial cannot be on a line with anything else

lnewson: It's in the wrong spot as I said in my last message.

modules/ROOT/pages/autocompleter.adoc
```diff
@@ -23,7 +23,12 @@ The two arguments this method take are:
 [cols="1,2,1,4",options="header"]
 |===
 |Name |Value |Requirement |Description
-|ch |string (of one character) |Required |The character to trigger the autocompleter.
+|ch |`+string+` (of one character) |Required |The character to trigger the autocompleter.
+include::partial$misc/admon-deprecated-6.2v.adoc[]
+It has been replaced with the `+trigger+` option, which should be used instead.
+|trigger |`+string+` |Required |The string to trigger the autocompleter.
+include::partial$misc/admon-requires-6.2v.adoc[]
+It replaces the deprecated `+ch+` option, and should be used instead of that option.
```
lnewson, line 31:
```suggestion
include::partial$misc/admon-requires-6.2v.adoc[]
```
You really don't need to be repeating yourself twice here.

modules/ROOT/pages/autocompleter.adoc
lnewson: Ahh right, you did have something for both which I missed sorry. That said you really don't need both and we don't normally have it for both so you can remove the `trigger` message as I left a suggest for.

MitchCTiny: If this current change doesn't fix what you are asking then I'm sorry I just don't understand. The first line of this option (and of the `trigger` option) describes its use, then the line after it has the deprecation (or requires) partial since that cannot go on the same line as anything else, and then the third line states the correct option to use. I put the final part after the partial since it makes more sense that way, as it is stating the replacement immediately after the deprecation (or requires) notice.

modules/ROOT/pages/autocompleter.adoc
MitchCTiny: I thought that was what you were asking in the other thread?

modules/ROOT/pages/autocompleter.adoc
```diff
@@ -23,7 +23,11 @@ The two arguments this method take are:
 [cols="1,2,1,4",options="header"]
 |===
 |Name |Value |Requirement |Description
-|ch |string (of one character) |Required |The character to trigger the autocompleter.
+|ch |`+string+` (of one character) |Required |The character to trigger the autocompleter.
+include::partial$misc/admon-deprecated-6.2v.adoc[]
+It has been replaced with the `+trigger+` option, which should be used instead.
```
ephox-mogran, line 28: Do we explain why we deprecate it anywhere? I assume in this case it's because autocompleters can now have triggers that are more than one character, but do we need to say that? Do we need to provide the "no-op" migration advice? Or is that just something we cover in Release Notes?

ephox-mogran (approved):

spocke (approved):

modules/ROOT/pages/autocompleter.adoc
lnewson: It's something that would normally go in the release notes if necessary from what I recall, as it ends up being too wordy to put here 🤷

MitchCTiny: I can confirm that this is in the in progress release notes @brian-forte

PR #2519, opened 2022-09-21
DOC-1803: Create `admon-requires-6.3v.adoc` partial for future release

MitchCTiny:
Related Ticket: DOC-1803

Description of Changes:
* Added `admon-requires-6.3v.adoc` as partial for future release

Pre-checks:

Review:


brian-forte:
This may be superstition on my part, but, over the past 30+ years I’ve had various systems tell me off for giving them text files without a new-line character as the last character in the file.

Literally thousands of *No newline at end of file* messages and warning (and the multiple variations thereof) has me almost reflexively inclined to add a hard-return to even one-line files.

But, this is a learned, almost Pavlovian response, with only a long history of wanting to avoid warnings in log files and at STD OUT as evidence for it being the right thing to do.

Hence this comment, rather than the edit I really want to make (for, perhaps, no good reason).

ephox-mogran:
@brian-forte the previous files also had a trailing newilne, so you can argue on the basis of consistency, as well. I know in some of our scripts for other projects we've been burnt by not having trailing newlines, because the last line of a list wasn't processed.

ephox-mogran (approved):

modules/ROOT/partials/misc/admon-requires-6.3v.adoc
```diff
@@ -0,0 +1 @@
+NOTE: This feature is only available for {productname} 6.3 and later.
```
brian-forte, line 1:
```suggestion
NOTE: This feature is only available for {productname} 6.3 and later.

```
Adding a blank line.

Consequent to @ephox-mogran’s [comment](https://github.com/tinymce/tinymce-docs/pull/2519#issuecomment-1254372895), I’m taking the leap from [commentary](https://github.com/tinymce/tinymce-docs/pull/2519#issuecomment-1253564668) to concrete suggestion.

tiny-ben-tran (approved):
Don't think this is scalable. Why can't we parameterised the version number?

MitchCTiny:
> Don't think this is scalable. Why can't we parameterised the version number?

Something to think about, I'm happy to replace all the current version partials with a new parameter and a single partial.

lnewson:
This **cannot** use variables as this needs to be hardcoded. Now we could have something where you need to specify variables to include this, but that is very messy. This is generally the least intrusive options and it's fairly scalable as you only need one per minor release.

brian-forte:
> > Don't think this is scalable. Why can't we parameterised the version number?
> 
> Something to think about, I'm happy to replace all the current version partials with a new parameter and a single partial.

A further something to think about: a single parameter would make some statements technically untrue.

Something that requires 6.1 will, with a single parameter, be presented as requiring 6.3 when 6.3 comes out.

MitchCTiny:
> > > Don't think this is scalable. Why can't we parameterised the version number?
> > 
> > 
> > Something to think about, I'm happy to replace all the current version partials with a new parameter and a single partial.
> 
> A further something to think about: a single parameter would make some statements technically untrue.
> 
> Something that requires 6.1 will, with a single parameter, be presented as requiring 6.3 when 6.3 comes out.

I meant a single parameter which is declared immediately before including a partial, then likely switched back to 6.0 after the partial, and by default. But I agree with above, this method is simple and won't cause confusion.

PR #2540, opened 2022-10-13
DOC-554: admon-inline-editing-mode-does-not-support-comments.adoc, customsidebar.adoc, introduction-to-tiny-comments.adoc, & use-tinymce-inline.adoc

brian-forte:
1. Added `admon-inine-editing-mode-does-not-support-comments.adoc` to `/partial/misc/`.

This file contains an IMPORTANT admonition noting, as the file-name suggests, that Comments are not supported when TinyMCE is run in inline editing mode.

2. Added includes to the above `partial` to `use-tinymce-inline.adoc` and `introduction-to-tiny-comments.adoc`.

3. Added a NOTE admonition to `customsidebar.adoc` noting that Sidebars do not work in inline editing mode.

Related Ticket: [DOC-554](https://ephocks.atlassian.net/browse/DOC-554)

Description of Changes:
* Comments (and Sidebars, generally) don’t work when TinyMCE is running in inline editing mode. Added documentation to this effect.

Pre-checks:

Review:


modules/ROOT/pages/customsidebar.adoc
```diff
@@ -6,6 +6,8 @@
 
 {productname} allows developers to create sidebars and add custom UI widgets inside a constrained and easily accessible area of the editor. The sidebar is designed to allow administrators and plugin developers to provide additional tools that can be accessed by {productname} users.
 
+NOTE: Sidebars do not work when {productname} is run in xref:use-tinymce-inline.adoc[inline editing mode].
```
lnewson, line 9: This isn't quite accurate... we should be saying it only works in classic/iframe mode as it also won't work for distraction free mode (though technically is a slimmed down inline mode).

This is also inconsistent with the opposite where we have https://github.com/tinymce/tinymce-docs/blob/staging/docs-6/modules/ROOT/partials/misc/admon-inline-only.adoc already. So I'd expect this to largely be the same as that just worded differently for classic/iframe mode 🤷 the linked one can of course be changed, the main thing is consistency.

brian-forte: To quote my commit message:

> Edited NOTE admonition in both files to be more explicit about what is and is not supported, and to make both admonitions consistent with each other.

TheSpyder (approved):

MitchCTiny (commented):
Now that `modules/ROOT/partials/misc/admon-iframe-only.adoc` exists, it might be appropriate to include that partial at the top of the `sidebar_show` option in this PR.

tiny-ben-tran (approved):

modules/ROOT/pages/customsidebar.adoc
```diff
@@ -6,6 +6,8 @@
 
 {productname} allows developers to create sidebars and add custom UI widgets inside a constrained and easily accessible area of the editor. The sidebar is designed to allow administrators and plugin developers to provide additional tools that can be accessed by {productname} users.
 
+include::partial$misc/admon-iframe-only.adoc[]
```
MitchCTiny, line 9: This partial uses the word option, where in this case a `sidebar` isn't an option, but a UI component which can be added through the `editor.ui.registry.addSidebar` api. I suggested that this partial be included under the `sidebar_show` option instead, but if it were to be included here then it may need to be changed to reflect the type of component used by the page calling it. Might be a bit too complicated though.

brian-forte: Corrected with [this commit](https://github.com/tinymce/tinymce-docs/pull/2540/commits/54b1f11b5b880997eef491d3fb504630c7a29f1c).

MitchCTiny (approved):

modules/ROOT/partials/misc/admon-inline-only.adoc
```diff
@@ -1 +1 @@
-NOTE: This option is only supported when using an `inline` editor. For more information on the differences between the editing modes, see xref:use-tinymce-inline.adoc[Inline editing mode].
\ No newline at end of file
+NOTE: This feature is only supported when {productname} is run in _inline_ mode. It is not supported in _iframe_ mode. For more information on the differences between the editing modes, see xref:use-tinymce-inline.adoc[Inline editing mode].
```
lnewson, line 1:
```suggestion
NOTE: This feature is only supported when {productname} is run in _inline_ mode. It is not supported in _classic_ mode. For more information on the differences between the editing modes, see xref:use-tinymce-inline.adoc[Inline editing mode].
```
Publicly it's called `classic` mode not iframe.

PR #2549, opened 2022-10-21
DOC-1845: tweak the 5.x supported versions table to improve readability

brian-forte:
Related Ticket: [DOC-1845](https://ephocks.atlassian.net/browse/DOC-1845)

Description of Changes:
* To improve readability the table columns displaying release and end-of-support dates are now left- or right-aligned, rather than centred.

Pre-checks:
- [x] Branch prefixed with `feature/5/` or `hotfix/5/`

Review:

[DOC-1845]: https://ephocks.atlassian.net/browse/DOC-1845?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/_includes/misc/supported_versions.md
+++ b/_includes/misc/supported_versions.md
@@ -2,21 +2,21 @@
 
 Supported versions of {{site.productname}}:
 
-| Version |   Release Date     |  End of Support   |
-|:-------:|:------------------:|:-----------------:|
-|   5.10  | October 20, 2021   |  April 20, 2023   |
-|   5.9   | September 15, 2021 |  March 15, 2023   |
-|   5.8   | May 19, 2021       | November 19, 2022 |
-|   5.7   | February 24, 2021  | August 24, 2022   |
-|   5.6   | December 8, 2020   |   June 8, 2022    |
-|   5.5   | October 21, 2020   |   April 21, 2022  |
-|   5.4   | July 23, 2020      | January 23, 2022  |
-|   5.3   | June 11, 2020      | December 11, 2021 |
-|   5.2   | March 9, 2020      | September 9, 2021 |
-|   5.1   | December 11, 2019  |  June 11, 2021    |
-|   5.0   | February 5, 2019   | August 5, 2020    |
-|   4.9   | November 27, 2018  | December 31, 2020 |
-|   4.8   |   July 11, 2018    | January 11, 2020  |
+| Version |       Release Date |     End of Support |
+|:--------|-------------------:|-------------------:|
+| 5.10    |   October 20, 2021 |     April 20, 2023 |
+| 5.9     | September 15, 2021 |     March 15, 2023 |
+| 5.8     |       May 19, 2021 |  November 19, 2022 |
+| 5.7     |  February 24, 2021 |    August 24, 2022 |
+| 5.6     |  December 08, 2020 |      June 08, 2022 |
+| 5.5     |   October 21, 2020 |     April 21, 2022 |
+| 5.4     |      July 23, 2020 |   January 23, 2022 |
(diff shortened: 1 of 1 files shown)


MitchCTiny (approved):

lnewson (changes requested):
This doesn't make this easier to read. At the very least the version should stay center aligned to me at least.

Here's the new version:
image
Here's the old version:
image

brian-forte:
> This doesn’t make this easier to read.

The key study on this is ‘The end of the line: a survey of unjustified typography’, by P Stiff, from *Information Design Journal*, 1996.

The paper isn’t freely available, unfortunately, but the key take-aways are:

1. Saccadic eye movemens are the key measure when it comes to trying to measure readability. Saccades are jerk-like eyeball movements which redirect the visual axis to a new point in space. The number of saccades and their duration is a proxy for how much work the eye is doing, which is a proxy for how much effort is required to read.

2. Fully justified text (flush-right and flush-left) produced less-accurate saccades when moving from the end of one line to the beginning of the next. However the time it took to correct these initial errors with a second saccade was comparable between fully-justified text and flush-left–ragged-right text.

3. Centred text, however, produced less-accurate saccades when moving from the end of one line to the beginning of the next compared to both fully-justified and flush-left–ragged-right text. It also showed sigificantly longer correction times for the second saccade and sometimes even a third saccade before the eye began tracking across the new line.

There was a small-scale follow-up study done by Microsoft’s Excel team back in the late-1990s with regards reading numbers and the evidence we have from this suggests

1. ragged-left–flush-right is easier to read (as measured by saccade accuracy and error-correction time) for numbers. However, this effect is completely lost if the positional notation is changed from line to line. That is, ragged-left–flush-right’s utility appears to be mostly in providing a fixed spot for each number’s value to be ascertained from.

So

```
   10.0
   1000
   4.35
```

is not a good approach.

But

```
      10.00
   1,000.00
       4.35
```

is

My table edits follow both bits of data.

The *Version* column treats the version strings as text (because they are: versions are ordinal numbers, which are about position or rank, not numeric value) and sets said text flush-left

And the dates columns are treated as numbers (because they are closer to being numbers in this circumstance)  and sets them flush-right.

> At the very least the version and titles should stay center aligned to me at least (as I’m assuming the idea was to keep dates aligned).

Putting heads out of alignment with the associated copy reduces the visual connection between a head and the associated copy.

ShiridiGandham:
@brian-forte I'm curious to know who requested this change? 

As far as I know 5.x is in maintenance mode and we don't usually make changes unless requested by either Product or CS. 

It looks like "readability " changes may require some CSS changes from DX/Design teams. Please check.

My take would be to put it back in backlog if it is not requested by product/CS.

lnewson:
One other thing I should have mentioned earlier sorry, is remember this content is meant to be read **left -> right** not **top -> bottom**, so having this vertically aligned like this and having the large horizontal gaps are the reason I'd consider this to be worse to read. Remember the only thing that is meant to be read top -> bottom is the version number and it's already fairly well grouped as we only have 2 or 3 numbers. If we had more, then vertically aligning it would make more sense.

TheSpyder (approved):
I believe this came up because the 5.10 EOL date had to be changed from `tbd` to the real date and in the process Brian realised this table was harder to read than the 6x equivalent:
https://www.tiny.cloud/docs/tinymce/6/support/#supported-tinymce-versions

This change isn't just about whether scanning vertically is something customers need to do. Having them aligned up in a way that the years line up makes them less visually noisy. Table rows in the 5x docs have alternate background shading so I don't see lining the version up with a date as a problem.

Most of these dates have expired so it's a bit of a moot point really.

PR #2555, opened 2022-11-01
DOC-1849: Improve documentation of issue object and a11ychecker_filter_issue option

MitchCTiny:
Related Ticket: DOC-1849

Description of Changes:
* Improve and move description of `issue` object
* Improve description of `a11ychecker_filter_issue` option
* Add and improve examples of `a11ychecker_filter_issue` option

Pre-checks:

Review:


modules/ROOT/pages/a11ychecker.adoc
```diff
@@ -393,3 +383,20 @@ console.log(issues);
   }
 ]
 ----
+
+[[issue]]
+=== The `+issue+` object
+
+Each `rule` in a11ychecker will generate an issue on any element which matches it. This issue is available as an object in a number of APIs, and contains the specific data used in the a11ychecker dialog, pertaining to the element which triggered it. With `a11ychecker_filter_issue`, these issues can be parsed before they are classified as issues, with the data provided in the `issue` object used in the callback function. 
+
+The `+issue+` object is provided to any function implementation of this option. This object provides relevant data pertaining to the triggered issue:
```
brian-forte, line 392:
```suggestion
The `+issue+` object is provided to any option that implements said option as a function. This object provides relevant data pertaining to the triggered issue:
```

_Function_ here reads like it should be in the adjectival form, _functional_; describing the form taken by the noun phrase, _implementation of this option_.

I suspect that’s not true, but it’s hard to parse this sentence without thinking the noun is masquerading as a comparative.

The suggestion above is likely wrong but it’s me grasping at re-phrasing this so the original noun phrase and the noun-that-looks-reads-an-adjective can both still be nouns.

The reason I suspect it’s not right is that the re-written noun phrase is now, strictly speaking, a verb phrase with a referent — _implements said option_.

As I said, grasping.

modules/ROOT/pages/a11ychecker.adoc
```diff
@@ -393,3 +383,20 @@ console.log(issues);
...
+
+* `+id+` : `+String+` A11ychecker issue identifier used by {productname}, such as D1, T4A. For a description and other details about the issue, see xref:accessibility-rules[Accessibility Rules].
+
+* `+description+` : `+String+` description of the issue, as will be displayed in the dialog.
+
+* `+severity+` : `+String+` severity level of the issue, either `+info+`, `+warning+`, or `+error+`.
```
brian-forte, line 398:
```suggestion
* `+description+` : `+String+` description of the issue; as will be displayed in the dialog.

* `+severity+` : `+String+` severity level of the issue; either `+info+`, `+warning+`, or `+error+`.
```

A semi-colon to delimit the two descriptive clauses is a stronger demarcator and encourages the reader to take a mental breath between each clause.

modules/ROOT/pages/a11ychecker.adoc
MitchCTiny:
```suggestion
The `+issue+` object provides relevant data pertaining to any issue generated by an element which violates an xref:a11ychecker.adoc#accessibility-rules[Accessibility rule]):
```
Ah, I missed this. I started writing this under the `a11ychecker_filter_issue` option, then found it already existed under the `getAudit()` API, so I moved the whole section to the bottom. It's an object which is generated in the audit (when the dialog is opened, or via the API), which is also provided to `a11ychecker_filter_issue` for filtering before the final audit is produced. Let me know if this sounds better.

modules/ROOT/pages/a11ychecker.adoc
MitchCTiny: Good idea, thanks

modules/ROOT/pages/a11ychecker.adoc
brian-forte: This reads cleanly and succinctly.

kemister85 (commented):
Updates for both .adoc files read in a clear and concise manner.

brian-forte (approved):

modules/ROOT/partials/configuration/a11ychecker_filter_issue.adoc
```diff
@@ -40,7 +40,26 @@ tinymce.init({
...
   a11ychecker_filter_issue: (issue) => {
-    return issue.element.nodeName.toLowerCase() !== 'table';
+	  return !(issue.element.nodeName.toLowerCase() === 'table' || issue.severity !== 'error');
+  }
+});
+----
+
+=== Example: Using `+a11ychecker_filter_issue+` to filter images with empty alternative text from the Accessibility Checker I1 rule
+
+The callback function in the following example will only return `false` for any issues with `'I1'` as the `+'id'+` image elements with an empty `+'alt+'` attribute, otherwise the issue won't be filtered out. This implementation can be useful as allowing images to have empty alternative text can be an alternative to applying the `+role="presentation"+` attribute to mark an image as `+decorative+`.
```
ephox-mogran, line 50: Is it a bit confusing to use `alternative` in terms of "another option" and `alternative` as in the name of the text in the same sentence?

ephox-mogran (approved):
I like the detail and contextual understanding you've provided.

modules/ROOT/partials/configuration/a11ychecker_filter_issue.adoc
```diff
@@ -3,7 +3,7 @@
 
 The `+a11ychecker_filter_issue+` option allows Accessibility Checker issues to be removed from the Accessibility Checker report using a callback function. This option is a flexible alternative to the `+a11ychecker_ignored_rules+` option. This option can remove issues from the results shown in the dialog and the `+getReport+` API.
 
-The function will be passed the results of the xref:a11ychecker.adoc#getreport[`+getReport()+` API]. To filter an issue from the Accessibility Checker report, the callback needs to return `+false+` for that particular issue.
+The function will be passed the results of the xref:a11ychecker.adoc#getreport[`+getReport()+` API]. To filter an issue from the Accessibility Checker report, the callback needs to return `+false+` for that particular issue. The function makes use of xref:a11ychecker.adoc#issue[`issue` objects] as the parameter for filtering issues; after they are generated by the audit, and before being displayed in the dialog.
```
TheSpyder, line 6: After clarifying "the results of the report api" I don't think the bit you added is necessary.

It's not clear to me whether this filter is applied during the `getReport` API function call or if integrators need to do that themselves when using the API. I didn't specify that, because I don't have time to check the code, but I think we should.

```suggestion
The callback function will be passed xref:a11ychecker.adoc#issue[`issue` objects] generated by the xref:a11ychecker.adoc#getreport[`+getReport()+` API]. To remove an issue from the Accessibility Checker report the callback needs to return `+false+` for that particular issue.
```

TheSpyder (approved):
Good change!

modules/ROOT/partials/configuration/a11ychecker_filter_issue.adoc
MitchCTiny: Good catch, I'll fix that.

modules/ROOT/partials/configuration/a11ychecker_filter_issue.adoc
MitchCTiny: Done, thanks

modules/ROOT/pages/a11ychecker.adoc
```diff
@@ -393,3 +383,20 @@ console.log(issues);
   }
 ]
 ----
+
+[[issue]]
+=== The `+issue+` object
+
+In a11ychecker, when the content within the editor is `audited`, each element is checked to ensure that no xref:a11ychecker.adoc#accessibility-rules[Accessibility rules] are violated. Any element which doesn't adhere to a rule will generate an `issue` within the audit; the details of which are to be displayed in the a11ychecker dialog.
```
brian-forte, line 390:
```suggestion
In a11ychecker, when the content within the editor is `audited`, each element is checked to ensure that no xref:a11ychecker.adoc#accessibility-rules[Accessibility rules] are violated. Any element which doesn’t adhere to a rule will generate an `issue` within the audit; the details of which are to be displayed in the a11ychecker dialog.
```

modules/ROOT/partials/configuration/a11ychecker_filter_issue.adoc
```diff
@@ -40,7 +40,26 @@ tinymce.init({
...
   a11ychecker_filter_issue: (issue) => {
-    return issue.element.nodeName.toLowerCase() !== 'table';
+	  return !(issue.element.nodeName.toLowerCase() === 'table' || issue.severity !== 'error');
+  }
+});
+----
+
+=== Example: Using `+a11ychecker_filter_issue+` to filter images with empty alternative text from the Accessibility Checker I1 rule
+
+The callback function in the following example will only return `false` for any issues with `'I1'` as the `+'id'+` image elements with an empty `+'alt+'` attribute, otherwise the issue won't be filtered out. This implementation can be useful as allowing images to have empty alternative text can be another method of applying the `+role="presentation"+` attribute to mark an image as `+decorative+`.
```
brian-forte, line 50:
```suggestion
The callback function in the following example will only return `false` for any issues with `'I1'` as the `+'id'+` image elements with an empty `+'alt+'` attribute, otherwise the issue won’t be filtered out. This implementation can be useful as allowing images to have empty alternative text can be another method of applying the `+role="presentation"+` attribute to mark an image as `+decorative+`.
```

PR #2625, opened 2023-02-12
DOC-1909: Re-write Supported Mobile Platforms section to match the Browser Compatibility section approach

brian-forte:
Ticket: DOC-1909, Re-write Supported Mobile Platforms section to match the Browser Compatibility section approach.

Changes:
* Re-written to fulfill the brief. This PR presents two suggested re-writes.
	* both use language that doesn’t require this page be constantly updated as new Android and iOS versions are released.
	* The first ties our support to the versions the two vendors — Google and Apple, respectively — support.
		* In particular, this ties our support to Google’s habit of supporting the current, penultimate and ante-penultimate versions of Android.
	* The second re-write sets our support to the current and penultimate versions (which, without explicitly saying so, apes Apple’s approach for iOS).
* Choosing which is to be published is now a key aspect of any Pull Request reviews and comments.

Pre-checks:
- [x] Changelog entry added

Review:

### DIFF
--- a/modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
+++ b/modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
@@ -1,30 +1,20 @@
 == Supported Mobile Platforms
 
-{productname} mobile is tested against, and is designed to run on the following mobile operating system and browser combinations:
+{productname} mobile is tested against, and is designed to run on, the default browser running on a currently vendor-supported version of both Android and iOS.
 
 [cols="1,1",options="header"]
 |===
-|Operating System |Browser
+|Operating System |Default browser
 
 |Android
 |Chrome
 
 |iOS
-| Safari
+|Safari
 |===
 
-{productname} supports the current and penultimate pairings of these combinations.
+Google supports the current and two previous versions of Android.
 
-////
-This matches Apple’s approach. They have long supported both the current and penultimate (previous) version of iOS.
+Apple supports the current and previous versions of iOS.
 
-However it does not match Google’s approach to Android.
-
-As of this note, Android 14 is the current version, but Google supports Android back to version 10.
-
-NB: Google’s support is not more generous than Apple’s, despite superficial appearances.
-
-In terms of what hardware remains supported by the software support approach used by both companies, Apple is ahead. Because Apple supports a much greater percentage of extant devices with each iOS update
-
-Google, by contrast, drops support for many more devices with OS updates.
-////
+NOTE: By default, both Android and iOS automatically update both the operating system and default browser. Consequently, {productname} mobile is only supported on the most recent releases of each supported operating system and browser combination.
--- a/modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
+++ b/modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
@@ -1,4 +1,4 @@
-== Supported Mobile Platforms (1)
+== Supported Mobile Platforms
 
 {productname} mobile is tested against, and is designed to run on, the default browser running on a currently vendor-supported version of both Android and iOS.
 
@@ -16,24 +16,5 @@
 Google supports the current and two previous versions of Android.
 
 Apple supports the current and previous versions of iOS.
-
-NOTE: By default, both Android and iOS automatically update both the operating system and default browser. Consequently, {productname} mobile is only supported on the most recent releases of each supported operating system and browser combination.
-
-
-
-== Supported Mobile Platforms (2)
-
-{productname} mobile is tested against, and is designed to run on, the default browser running on a the current and previous versions of both Android and iOS.
-
-[cols="1,1",options="header"]
-|===
-|Operating System |Default browser
-
-|Android
-|Chrome
-
-|iOS
-|Safari
-|===
 
 NOTE: By default, both Android and iOS automatically update both the operating system and default browser. Consequently, {productname} mobile is only supported on the most recent releases of each supported operating system and browser combination.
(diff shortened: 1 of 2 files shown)


kemister85 (approved):

modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
```diff
@@ -1,12 +1,39 @@
...
+|Android
+|Chrome
+
+|iOS
+|Safari
 |===
+
+Google supports the current and two previous versions of Android.
+
+Apple supports the current and previous versions of iOS.
```
mkzhx, line 18: How many previous versions of iOS? Is it known/consistent?

modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
```diff
@@ -1,12 +1,39 @@
...
+
+NOTE: By default, both Android and iOS automatically update both the operating system and default browser. Consequently, {productname} mobile is only supported on the most recent releases of each supported operating system and browser combination.
+
+
+
+== Supported Mobile Platforms (2)
```
mkzhx, line 24: Why is there a duplicate section?

modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
```diff
@@ -1,12 +1,39 @@
-== Supported Mobile Platforms
+== Supported Mobile Platforms (1)
```
mkzhx, line 1: Remove ` (1)`

modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
brian-forte: > Why is there a duplicate section

See the [description](https://github.com/tinymce/tinymce-docs/pull/2625#issue-1581238819).

modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
brian-forte: The out-of-date version of this section (and previous versions of this section) always listed the then-current and then-penultimate versions of both iOS and Android.

Version (2) of this section preserves this but re-writes it such that the section does not need constant updating.

Version (1), by contrast, has us supporting TinyMCE on all vendor-supported versions of Android and iOS.

The vendor-supported versions are the current and penultimate versions of iOS (the two versions Apple supports) and the current, penultimate and ante-penultimate versions of Android (the three versions Google supports).

Version (2) is the status quo re-phrased to no longer require constant updating.

Version (1) is a change that would require confirmation of this actually being the case. But would also bring how we support TinyMCE on Android in line with how Google supports Android on mobile devices.

modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
brian-forte: Once one section gets all the appropriate confirmations and blessings, the identifying sub-string in the section that goes into production will have the identifier removed.

modules/ROOT/partials/misc/mobile-platform-compatibility.adoc
mkzhx: Is removing the "Google supports..." and "Apple supports..." lines, but preserving the phrase "vendor-supported" a good compromise? The assumption is that people will know or research which versions are vendor-supported themselves.

brian-forte: >Is removing the “Google supports…” and “Apple supports…” lines, but preserving the phrase “vendor-supported” a good compromise? The assumption is that people will know or research which versions are vendor-supported themselves.

Simpler is a good argument for removing the lines, but it still requires confirmation that we do, in fact, support Android through all three versions Google supports, rather than the two (current and penultimate) we always, previously, said we supported.

FWIW, the reason for bringing up the notion of matching our support with what the venders supported is that vendor-support is commonly a measure used by companies for what they, themselves, support.

If a company deploys a mobile app which incorporates TinyMCE and they have Android devices in staff hands that are running what is, at the time of deployment, the ante-penultimate version of Android, our not supporting TinyMCE on that device while Google still does support said device might not sit well with said company.

mkzhx: With those two lines removed I still believe it is an improvement over version 2 though. Regardless, I support version 1 even if it means we have to ensure that the two lines are up to date - I think if it ever changes it will be big enough news that we don't have to explicitly search for and keep track of it.

mkzhx (approved):

TheSpyder (approved):
Option 1 is my preference, and expanding to "every version google supports" is ok (but you should check whether QA is comfortable with us saying that).

As I read this I'm also realising why we had the specific version language before. In cases where we support two major versions of TinyMCE simultaneously, we won't update the older release to fix mobile bugs in newer operating system versions. The old language implied that by not listing the new OS version.

I think this change is still fine, however, as despite the new wording nothing will actually require us to update old versions of TinyMCE for mobile OS updates.

brian-forte:
@TheSpyder,

> As I read this I’m also realising why we had the specific version
> language before. In cases where we support two major versions of
> TinyMCE simultaneously, we won’t update the older release to fix
> mobile bugs in newer operating system versions. The old language
> implied that by not listing the new OS version.

It’s a bit cludgy but I think we are covered on this front by the 5.x docs [still having version numbers listed in the mobile chapter](https://tiny.cloud/docs/mobile/). A happy if accidental consequence of the 5.x docs not being actively updated any longer.

> you should check whether QA is comfortable with us saying that

I’ll push a _ready-to-merge_ update and ping QA folk, asking them to chime in.

kemister85 (approved):

MitchCTiny (approved):

PR #2637, opened 2023-03-01
removed paste plugin references

nynevi:
Ticket: No ticket, don't know how to open a ticket, could not find it in the Contribution guidelines

Changes:
* Removes `paste` plugin references from all the documentation examples
* Adds a note to the `paste` plugin page that it has been part of the core package since 6.0.0

Pre-checks:
- [ ] Branch prefixed with `feature/6/` or `hotfix/6/`
- [ ] Changelog entry added
- [ ] `modules/ROOT/nav.adoc` has been updated (if applicable)

I did not tick any checkboxes above as it is neither a feature nor a hotfix and the changelog of 6.0.0 already includes the plugins being moved to core package. The documentation should have been updated on version 6.0.0 2 years ago when the change to move paste into the core was done.

Here are list of tasks that are just closed while the documentation stayed as it is, misleading:
- https://github.com/tinymce/tinymce/issues/7685
- https://github.com/tinymce/tinymce/issues/8378

Review:

### DIFF
--- a/_includes/configuration/help_tabs.md
+++ b/_includes/configuration/help_tabs.md
@@ -15,7 +15,7 @@ If `help_tabs` is not configured, any tabs defined using `addTab` will be displa
 ```js
 tinymce.init({
   selector: 'textarea',
-  plugins: 'help link table paste code emoticons',
+  plugins: 'help link table code emoticons',
   toolbar: 'help addTab',
   help_tabs: [
     'shortcuts', // the default shortcuts tab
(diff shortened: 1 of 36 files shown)


TheSpyder:
I think you might be confused as we published v6 docs using a new system and haven't migrated the old documentation to it.

The `main` branch is for [TinyMCE 5 docs](https://www.tiny.cloud/docs/), these include the paste plugin because so does that version. The tickets we closed were logged in the gap between community and enterprise release, at which point the [TinyMCE 6 docs](https://www.tiny.cloud/docs/tinymce/6/) were published.

If you look at the `release/docs-6` branch there is no reference to the paste plugin (other than the migration guide which says we removed it).

As a side note, TinyMCE 6 was released 1 year ago (3 march 2022), not 2 years ago.

nynevi:
Makes sense, tha ks

PR #2757, opened 2023-05-25
DOC-2029: Document new editable root APIs and options

spocke:
Ticket: DOC-2029, Document new editable root APIs and options.

Changes:
* Added docs for the new `editable_root` and `newdocument_content` options unsure where to stick them none of the categories seems right to me. So I put both under `content-behavior-options.adoc`.

Pre-checks:
- [x] Changelog entry added
~- [ ] `modules/ROOT/nav.adoc` has been updated (if applicable)~
~- [ ] Files removed have been deleted, not just excluded from the build (if applicable)~

Review:

### DIFF
--- a/modules/ROOT/partials/configuration/editable_root.adoc
+++ b/modules/ROOT/partials/configuration/editable_root.adoc
@@ -1,0 +1,20 @@
+[[editable_root]]
+== `+editable_root+`
+
+This option enables you to specify that the initial editable state of the root of {productname}. The root could still contain child elements with `contenteditable="true"` and those
+would remain editable even if you set this option to `false`.
+
+*Type:* `+Boolean+`
+
+*Default value:* `+'true'+`
+
+=== Example: using `+editable_root+`
+
+[source,js]
+----
+tinymce.init({
+  selector: 'textarea',  // change this value according to your HTML
+  editable_root: false 
+});
+----
+
--- a/modules/ROOT/partials/configuration/editable_root.adoc
+++ b/modules/ROOT/partials/configuration/editable_root.adoc
@@ -1,12 +1,15 @@
 [[editable_root]]
 == `+editable_root+`
 
-This option enables you to specify that the initial editable state of the root of {productname}. The root could still contain child elements with `contenteditable="true"` and those
-would remain editable even if you set this option to `false`.
+This option sets the initial editable state of a {productname} instance’s root.
+
+Setting `editable_root` to `false` sets the {productname} root to non-editable (in effect, read-only).
+
+NOTE: a {productname} root set as not-editable can contain child elements with the `contenteditable="true"` attribute set. And these child elements remain editable even when `editable_root: false` is set.
 
 *Type:* `+Boolean+`
 
-*Default value:* `+'true'+`
+*Default value:* `+true+`
 
 === Example: using `+editable_root+`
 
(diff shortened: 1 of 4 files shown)


modules/ROOT/partials/configuration/editable_root.adoc
```diff
@@ -0,0 +1,20 @@
+[[editable_root]]
+== `+editable_root+`
+
+This option enables you to specify that the initial editable state of the root of {productname}. The root could still contain child elements with `contenteditable="true"` and those
+would remain editable even if you set this option to `false`.
```
MitchCTiny, line 5:
```suggestion
This option enables you to specify that the initial editable state of the root of {productname}.
Child elements with the `contenteditable="true"` attribute will still be editable when this option is `false`.
```

modules/ROOT/partials/configuration/newdocument_content.adoc
```diff
@@ -0,0 +1,21 @@
...
+*Default value:* `+''+`
+
+=== Example: using `+newdocument_content+`
+
+[source,js]
+----
+tinymce.init({
+  selector: 'textarea',  // change this value according to your HTML
+  newdocument_content: '<p contenteditable="true">Editable content</p>' 
+});
```
MitchCTiny, line 18:
```suggestion
This option enables you to set the content that the editor will contain if you use the New document menu item.

*Type:* `+String+`

*Default value:* `+''+`

=== Example: using `+newdocument_content+`

This example shows `+newdocument_content+` used together with the xref:content-behavior-options.adoc#editable_root[`+editable_root+` option] to ensure that there is always an editable element within the non-editable root.

[source,js]
----
tinymce.init({
  selector: 'textarea',  // change this value according to your HTML
  editable_root: false,
  newdocument_content: '<p contenteditable="true">Editable content</p>' 
});
```
Moved the use case next to the relevant example. There should also be a link connected to "New document menu item", to a relevant section of the docs. And perhaps this option should be included there as well?

brian-forte, line 18:
```suggestion
This option sets the content a new editor contains when the xref:available-menu-items.adoc#the-core-menu-items[File -> New document] menu item is invoked.

*Type:* `+String+`

*Default value:* `+''+`

=== Example: using `+newdocument_content+`

This example shows `+newdocument_content+` being used with the xref:content-behavior-options.adoc#editable_root[`+editable_root+`] option to set an editable element within a non-editable root.

[source,js]
----
tinymce.init({
  selector: 'textarea',  // change this value according to your HTML
  editable_root: false,
  newdocument_content: '<div contenteditable="true">Editable content</div>' 
});
```
1. Copy-edits and edits.
2. Incorporates @MitchCTiny’s structural edit above.
3. Also adds an xref as per @MitchCTiny’s editorial note.
4. Takes a liberty with the example and changes the tag set in place by`newdocument_content` tag `<p>` to `<div>`.
	* Thinking like the sys-admin I once was for a moment, I’d put my _end-users can edit this part of the form_ chunks inside `<div></div>` tags rather than <p></p> tags.

spocke: All makes sense thanks for these changed updated the PR.

MitchCTiny (approved):

PR #3455, opened 2024-09-26
DOC-2534: Update the example for lazy loading a revision

tiny-ben-tran:
Ticket: DOC-2534

Site: [Staging branch](http://docs-<hotfix|feature>-<version>-doc-<num>.staging.tiny.cloud/docs/tinymce/<version>/)

Changes:
* Updated the example to throw an error when the revision being fetched is not found 
* Updated the example.js of full feature to correctly reflect the revisions being shown

Pre-checks:
- [x] `modules/ROOT/nav.adoc` has been updated `(if applicable)`.
- [x] ~Included a `release note` entry for any `New product features`.~
- [x] If this is a minor release, updated `productminorversion` in `antora.yml` and added new supported versions entry in `modules/ROOT/partials/misc/supported-versions.adoc`.

Review:

### DIFF
--- a/modules/ROOT/examples/live-demos/revisionhistory/index.js
+++ b/modules/ROOT/examples/live-demos/revisionhistory/index.js
@@ -185,15 +185,12 @@ const revisions = [
 const revisionhistory_fetch_revision = (_editor, revision) =>
   new Promise((resolve, reject) => {
     setTimeout(() => {
-      for (let i = 0; i < revisions.length; i++) {
-        const temp = revisions[i];
-        if (temp.revisionId === revision.revisionId) {
-          resolve(temp);
-          break;
-        }
+      const newRevision = revisions.find((r) => r.revisionId === revision.revisionId);
+      if (newRevision === undefined) {
+        reject(`Revision ${revision.revisionId} is not found`);
+      } else {
+        resolve(newRevision);
       }
-      // reject has no effect when the promise is resolved
-      reject(`Revision ${revision.revisionId} is not found`);
     }, getRandomDelay());
   });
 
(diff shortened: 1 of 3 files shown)


modules/ROOT/examples/live-demos/revisionhistory/index.js
```diff
@@ -183,17 +183,17 @@ const revisions = [
 ];
 
 const revisionhistory_fetch_revision = (_editor, revision) =>
-  new Promise((resolve) => {
+  new Promise((resolve, reject) => {
     setTimeout(() => {
-      let newRevision = null;
       for (let i = 0; i < revisions.length; i++) {
```
shanmen-tiny, line 188:
```suggestion
      const newRevision = revisions.find((r) => r.revisionId === revision.revisionId);
      if (newRevision === undefined) {
    	reject(`Revision ${revision.revisionId} is not found`);
      } 
      resolve(newRevision);
```
Just a suggestion 👍

shanmen-tiny (approved):

modules/ROOT/examples/live-demos/revisionhistory/index.js
```diff
@@ -183,17 +183,17 @@ const revisions = [
...
         const temp = revisions[i];
         if (temp.revisionId === revision.revisionId) {
-          newRevision = temp;
+          resolve(temp);
           break;
         }
       }
-      resolve(newRevision);
+      // reject has no effect when the promise is resolved
```
TheSpyder, line 195: I appreciate that this is how promises work, but I would still like to encourage better coding practices - particularly for our code examples we expect customers to use (and often deploy to production).

That means either using an `else` block or switching to an `async` function and early `return`.

modules/ROOT/examples/live-demos/revisionhistory/index.js
tiny-ben-tran: This is acceptable because IE is no longer supported :)

kemister85 (approved):

PR #3717, opened 2025-05-06
DOC-3151: Changed word-capture regex for abbreviations to allow larger abbreviations.

abhinavgandham:
Ticket: DOC-3151

Site: [Staging branch](http://docs-feature-790-doc-3151tiny-11560.staging.tiny.cloud/docs/tinymce/latest/7.9.0-release-notes/#changed-word-capture-regular-expression-for-abbreviations-to-allow-larger-abbreviations)

Changes:
* Documented the bug fix for TINY-11560.

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/pages/7.9.0-release-notes.adoc
+++ b/modules/ROOT/pages/7.9.0-release-notes.adoc
@@ -170,7 +170,7 @@ This behavior was caused by insufficient overlap detection when identifying matc
 <p>D. dd.DDD.</p>
 ----
 
-.Example after the fix
+.Example after the fix:
 [source,html]
 ----
 <!-- Original content typed by the user -->
--- a/modules/ROOT/pages/7.9.0-release-notes.adoc
+++ b/modules/ROOT/pages/7.9.0-release-notes.adoc
@@ -153,11 +153,12 @@ For information on using Enhanced Skins & Icon Packs, see: xref:enhanced-skins-a
 
 {productname} {release-version} also includes the following bug fix<es>:
 
-// === <TINY-vwxyz 1 changelog entry>
-// #TINY-vwxyz1
+=== Changed word-capture regular expression for abbreviations to allow larger abbreviations.
+// #TINY-11560
 
-// CCFR here.
+Previously in {productname}, an issue was identified where words or letters were inadvertently duplicated during pattern matching. This behavior particularly affected the handling of acronyms, resulting in inconsistent behavior.
 
+With the release of {productname} {release-version}, this issue has been resolved. Pattern matching now performs as expected, ensuring text is processed accurately without unintended duplication.
 
 [[security-fixes]]
 == Security fixes
--- a/modules/ROOT/pages/7.9.0-release-notes.adoc
+++ b/modules/ROOT/pages/7.9.0-release-notes.adoc
@@ -153,12 +153,32 @@ For information on using Enhanced Skins & Icon Packs, see: xref:enhanced-skins-a
 
 {productname} {release-version} also includes the following bug fix<es>:
 
-=== Changed word-capture regular expression for abbreviations to allow larger abbreviations.
+=== Improved handling of overlapping matches in abbreviation detection
 // #TINY-11560
 
-Previously in {productname}, an issue was identified where words or letters were inadvertently duplicated during pattern matching. This behavior particularly affected the handling of acronyms, resulting in inconsistent behavior.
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/7.9.0-release-notes.adoc
```diff
@@ -153,11 +153,12 @@ For information on using Enhanced Skins & Icon Packs, see: xref:enhanced-skins-a
 
 {productname} {release-version} also includes the following bug fix<es>:
 
-// === <TINY-vwxyz 1 changelog entry>
-// #TINY-vwxyz1
+=== Changed word-capture regular expression for abbreviations to allow larger abbreviations.
+// #TINY-11560
 
-// CCFR here.
+Previously in {productname}, an issue was identified where words or letters were inadvertently duplicated during pattern matching. This behavior particularly affected the handling of acronyms, resulting in inconsistent behavior.
```
HAFRMO, line 159: If you want some details, the problem was previously that in some situations we got matching patterns that'd overlap and cause the situation such as:
For the text `D. dd.D.` you'd get patterns that matched `D.` and `dd.D`. When pattern-matching what we should do was to match each once and ignored the overlap on the last 'D'. We didn't and instead got a pattern-matching the last 'D' twice. In the process of pattern-matching we'd remove the content from the editor, then insert it back ( I don't know why ). As we'd found the patterns `D.`, ( Space wasn't removed, didn't need to be inserted back ) `dd.D`, AND `D.` we'd naturally insert them back again, resulting in the text `D. dd.DD.`. ( It actually became `D. dd.DDD.` due to a different piece of code which assumed that there was no overlapping regions actually ended up adding the overlapping area as its own detected piece, so we actually 'found' the patterns `D.`, `dd.D`, `D`, and `D.` ).
The fix was just to improve area overlap detection, so we only detect `D.` and `dd.D`. This meant that there was no overlap.
And yes, the first `D. ` was just to trigger the server's spellcheck to add the pattern for it, it doesn't actually involve itself with the problem. `D. bunch of filler text dd.D.` would have resulted in `D. bunch of filler text dd.DDD.` too. It makes the explanation a bit weirder ( due to the space which isn't detected but kept ) but that's how it is.

HAFRMO (approved):

modules/ROOT/pages/7.9.0-release-notes.adoc
kemister85: @HAFRMO For your review

I do have one question, why do we now have `languageMismatchValidation: true` during the check now? I dont see any mention or implementation for this new boolean field in changes.

Screenshot 2025-05-06 at 9 42 47 pm


=== Improved handling of overlapping matches in abbreviation detection
// #TINY-11560

Previously in {productname}, an issue occurred during pattern matching where overlapping regular expression matches caused duplicated content when processing abbreviations. For example, in the string `D. dd.D.`, multiple overlapping patterns such as `D.`, `dd.D`, and a final `D.` could be incorrectly matched and processed more than once. This resulted in unexpected duplication in the editor, such as `D. dd.DDD.`.

This behavior was caused by insufficient overlap detection when identifying match regions, leading to multiple reinsertions of already matched substrings. The issue has now been resolved in {productname} {release-version} by refining the pattern-matching logic to accurately detect and avoid overlapping matches. The updated implementation ensures that each abbreviation is matched and processed only once, eliminating duplicated insertions and preserving the original content structure.

.Example before the fix:
[source,html]
----

<p>D. dd.D.</p>


<p>D. dd.DDD.</p>
----

.Example after the fix
[source,html]
----

<p>D. dd.D.</p>


<p>D. dd.D.</p>
----

kemister85: @abhinavgandham Lets go with the above suggestion content change.

abhinavgandham: @kemister85 yeah agreed

modules/ROOT/pages/7.9.0-release-notes.adoc
```diff
@@ -153,11 +153,32 @@ For information on using Enhanced Skins & Icon Packs, see: xref:enhanced-skins-a
...
+[source,html]
+----
+<!-- Original content typed by the user -->
+<p>D. dd.D.</p>
+
+<!-- Result after pattern matching (incorrect behavior) -->
+<p>D. dd.DDD.</p>
+----
+
+.Example after the fix
```
abhinavgandham, line 173:
```suggestion
.Example after the fix:
```

modules/ROOT/pages/7.9.0-release-notes.adoc
abhinavgandham: @kemister85 The changes have been applied.

ShiridiGandham (approved):
LGTM

PR #3895, opened 2025-10-27
DOC-3310: Fix `uploadcare_resources` option examples

MitchC1999:
Ticket: DOC-3310

Site: [Staging branch](http://docs-hotfix-8-doc-3310.staging.tiny.cloud/docs/tinymce/latest/uploadcare/)

Changes:
* Fix examples for `uploadcare_resources` option
* Improve wording

Pre-checks:
- [x] `modules/ROOT/nav.adoc` has been updated `(if applicable)`.
- [x] Included a `release note` entry for any `New product features`.
- [x] If this is a minor release, updated `productminorversion` in `antora.yml` and added new supported versions entry in `modules/ROOT/partials/misc/supported-versions.adoc`.

Review:

### DIFF
--- a/modules/ROOT/partials/configuration/uploadcare_resources.adoc
+++ b/modules/ROOT/partials/configuration/uploadcare_resources.adoc
@@ -20,9 +20,9 @@ tinymce.init({
 });
 ----
 
-=== Directory Structure
+=== Resources Structure
 
-When using `uploadcare_resources`, ensure your directory structure follows this pattern, specifically the `js` and `css` subdirectories:
+When using `uploadcare_resources`, ensure your directory structure follows this pattern, specifically the `js` and `css` subdirectories, and that the necessary resource files are present:
 
 .Example directory structure:
 [source,text]
--- a/modules/ROOT/partials/configuration/uploadcare_resources.adoc
+++ b/modules/ROOT/partials/configuration/uploadcare_resources.adoc
@@ -1,18 +1,11 @@
 [[uploadcare-resources]]
 == `uploadcare_resources`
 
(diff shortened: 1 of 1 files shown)


kemister85 (approved):

modules/ROOT/partials/configuration/uploadcare_resources.adoc
```diff
@@ -22,53 +15,19 @@ tinymce.init({
   plugins: 'uploadcare',
   toolbar: 'uploadcare uploadcare-video',
   uploadcare_public_key: '<your-public-key>',
-  uploadcare_resources: 'uploadcare/uploadcare', // Points to the uploadcare plugin directory
+  uploadcare_resources: 'path/to/uploadcare-resources/',
```
lorenzo-pomili, line 18:
```suggestion
  uploadcare_resources: 'path/to/uploadcare-plugin/',
```
I think saying that is the path to the plugin is more clear

lorenzo-pomili (approved):

modules/ROOT/partials/configuration/uploadcare_resources.adoc
MitchC1999: This option only concerns the resources and would only really be necessary when the resources are in a separate folder to the plugin. The default value is the plugin directory, so if it was set to the same folder then you wouldn't even need to specify the option.

lorenzo-pomili: no, wait this option is needed to get the resources, but it needs to point at the builded plugin path since the resources will be in that folder

MitchC1999: Yes that's where they are in the build, but the integrator can put them somewhere else, apart from the plugin, and use this option to load them from there.

PR #3932, opened 2025-12-01
DOC-3224:  `powerpaste_emf_processor` function that takes an emf blob and should return a promise of a png blob.

abhinavgandham:
Ticket: DOC-3224

Site: [Release note entry](https://pr-3932.tinymce-docs.iad.staging.tiny.cloud/docs/tinymce/latest/8.3.0-release-notes/#powerpaste_emf_processor-function-that-takes-an-emf-blob-and-returns-a-promise-of-a-png-blob)

Site: [Option documentation](https://pr-3932.tinymce-docs.iad.staging.tiny.cloud/docs/tinymce/latest/powerpaste-options/#powerpaste_emf_processor)

Changes:
* Documented the new `powerpaste_emf_processor` option.

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/pages/8.3.0-release-notes.adoc
+++ b/modules/ROOT/pages/8.3.0-release-notes.adoc
@@ -40,7 +40,7 @@ Before October 2025, copying and pasting content from Microsoft Word containing 
 
 This change negatively affected the user experience. After October 2025, users would see a low-quality image alongside an error message stating `Some images failed to import`, which caused confusion since the image was, in fact, visible in the editor—albeit with degraded quality.
 
-{productname} {release-version} introduces a new option, xref:powerpaste_emf_processor.adoc#powerpaste_emf_processor[powerpaste_emf_processor], to address this issue. This option accepts a function that receives an `emf` blob and returns a promise resolving to a `png` blob (i.e. `(blob: Blob) => Promise<Blob>`).
+{productname} {release-version} introduces a new option, xref:powerpaste-options.adoc#powerpaste_emf_processor[powerpaste_emf_processor], to address this issue. This option accepts a function that receives an `emf` blob and returns a promise resolving to a `png` blob (i.e. `(blob: Blob) => Promise<Blob>`).
 
 * If this option is provided, the editor will invoke the supplied function to convert `emf` image blobs into `png` image blobs before inserting them into the editor.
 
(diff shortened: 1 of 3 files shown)


modules/ROOT/pages/8.3.0-release-notes.adoc
hamza0867, line 1: It is I true I mentioned the difference started around October 2025, but we should tell the customers that a release of Word made a change to the clipboard data they provide us. So the cause of the difference in behavior is a change introduced in a recent release of Word, not the month October 2025. Does that make sense?

We should also pinpoint the version that introduced this breaking change. I know for Macs, it's somewhere between 16.98 and 16.102 but I don't know which one exactly introduced the breaking change

abhinavgandham: @hamza0867 Noted, would you happen to know any resource that I can use to find the specific version that introduced the breaking change?

metricjs: Chiming in - I can't find a useful changelog for Word that would tell us which version (especially as they often say "various fixes") and testing manually seems like it would require a lot of work. Unless we have an easy way to figure it out, I'd almost be tempted to put "around version 16.100" or something - or even the version range you gave above.

hamza0867: Sorry I missed this comment.

Yes, @metricjs is correct here. The only way to know for sure is to download each version and test it and that is by no means easy to do. So I say let's go with `a version released in October`. What do you think ?

metricjs: Let's make sure we say "October 2025" just to be specific, but yes, agreed.

metricjs: Which I believe means the current content is good as is?

kemister85 (approved):
LGTM, sounds like everyone is on the same page also which is good.

hamza0867 (approved):

modules/ROOT/partials/configuration/powerpaste_emf_processor.adoc
```diff
@@ -0,0 +1,22 @@
+[[powerpaste_emf_processor]]
+== `+powerpaste_emf_processor+`
+
+This option controls how `emf` images are handled. If this option **is** provided, `emf` image blobs will be converted into `png` image blobs, which will then be pasted into the editor. If this option is **not** provided, the editor will remove the poor quality images, and show an error message.
```
metricjs:
```suggestion
This option controls how `emf` images are handled. If this option **is** provided, `emf` image blobs will be converted into `png` image blobs using the supplied function, which will then be pasted into the editor. If this option is **not** provided, the editor will remove the poor quality images, and show an error message.
```
nitpick, but figured it doesn't hurt to be explicit

metricjs (approved):

PR #3933, opened 2025-12-02
DOC-3224: Show editor notification when a premium plugin is not allowed for a given license key.

abhinavgandham:
Ticket: DOC-3224

Release Note Entry: [Staging branch](https://pr-3933.tinymce-docs.iad.staging.tiny.cloud/docs/tinymce/latest/8.3.0-release-notes/#show-editor-notification-when-a-premium-plugin-is-not-allowed-for-a-given-license-key)

Updated License Key error messages table: [Staging branch](https://pr-3933.tinymce-docs.iad.staging.tiny.cloud/docs/tinymce/latest/license-key/#license-key-error-messages)

Changes:
* Documented the license key improvement for TINY-12937. Updated the license key error message table to now contain the message `One or more premium plugins are disabled due to license key restrictions.` for the `invalid plugin` case.

Pre-checks:

Review:

### DIFF
--- a/modules/ROOT/pages/8.3.0-release-notes.adoc
+++ b/modules/ROOT/pages/8.3.0-release-notes.adoc
@@ -46,10 +46,12 @@ The following premium plugin updates were released alongside {productname} {rele
 
 {productname} {release-version} also includes the following improvements:
 
-=== <TINY-vwxyz 1 changelog entry>
-// #TINY-vwxyz1
+=== Show editor notification when a premium plugin is not allowed for a given license key.
+// #TINY-12937
 
-// CCFR here.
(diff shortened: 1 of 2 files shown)


metricjs (approved):
I'm happy with the content as is, however we could optionally expand it to note that:
- The details of which plugins are not allowed can be found in the console
- The error message will be the same regardless of if one or multiple plugins are not allowed

Up to you.

abhinavgandham:
@metricjs I think its a good idea to do that, I can just add a note below the release note entry with that information. Does this sound good?

kemister85 (approved):

PR #4028, opened 2026-03-16
DOC-3386: Fix dependency vulnerabilities via http-server upgrade and minimal resolutions

kemister85:
Ticket: DOC-3386

Changes:
- Upgrade http-server from ^0.12.3 to ^14.1.1 (v14 no longer uses ecstatic)
- Remove ecstatic resolution and direct devDep
- Keep liquidjs resolution (antora-extension-livedemos pins ^9.37.0; fix requires updating that package)

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,15 +28,7 @@
     "ext": "*"
   },
   "resolutions": {
-    "brace-expansion": ">=1.1.12",
-    "sha.js": ">=2.4.12",
-    "lodash": ">=4.17.23",
-    "js-yaml": ">=4.1.1",
-    "qs": ">=6.14.2",
-    "minimatch": ">=3.1.4",
-    "ecstatic": ">=4.1.3",
-    "liquidjs": ">=10.25.0",
-    "yauzl": ">=3.2.1"
+    "liquidjs": ">=10.25.0"
   },
   "devDependencies": {
     "@antora/cli": "^3.1.10",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -27,6 +27,17 @@
...
+    "brace-expansion": ">=1.1.12",
+    "sha.js": ">=2.4.12",
+    "lodash": ">=4.17.23",
+    "js-yaml": ">=4.1.1",
+    "qs": ">=6.14.2",
+    "minimatch": ">=3.1.4",
+    "ecstatic": ">=4.1.3",
+    "liquidjs": ">=10.25.0",
+    "yauzl": ">=3.2.1"
+  },
```
TheSpyder, line 32: Wouldn't it be better to upgrade the direct dependencies so the transitive ones are correct?

This approach can lead to bugs when transitive dependencies are upgraded to new major versions.

kemister85: Good point. I’ve switched to upgrading the direct dependency instead of using resolutions where possible.

Changes:

- Upgraded http-server from ^0.12.3 to ^14.1.1 – v14 no longer uses ecstatic, so the ecstatic resolution and direct devDep are removed.
- Removed the other resolutions (brace-expansion, sha.js, lodash, js-yaml, qs, minimatch, yauzl) – they’re no longer needed; upstream fixes cover them.
- Kept the liquidjs resolution – @tinymce/antora-extension-livedemos still pins liquidjs ^9.37.0. Fixing that would require updating that package and publishing a new version.

Result from local testing:
Build and serve both work, and yarn audit reports 0 vulnerabilities.

kemister85: @TheSpyder Thoughts?

kemister85:
Closing this now, as this was resolved by https://github.com/tinymce/tinymce-docs/pull/4130

PR #4260, opened 2026-07-07
TINYDOC-3531 - Document expand and split format parameters for content formatting

kemister85:
**Ticket:** TINYDOC-3531

**Site:** [Staging branch](https://pr-4260.tinymce-docs.iad.staging.tiny.cloud/docs/tinymce/latest/content-formatting/#format-parameters)

**Changes:**
* Documented the `expand` format parameter in `modules/ROOT/partials/configuration/formats.adoc` (Format parameters section): description, type, default (`true` for selector formats), possible values, and an example. `expand` controls whether the range is expanded to cover the whole matching selector element before a format is applied or removed; `expand: false` restricts the operation to the current selection.
* Documented the `split` format parameter: description, type, default (`true` for inline/non-selector formats, `false` for selector/block formats), possible values, and an example. `split` controls whether the wrapping element is split at the selection boundaries when a format is removed.
* Both parameters were already used in the existing `block_expand`, `deep`, and `removeformat` examples on this page but were undefined; they are now listed alongside the other format object properties.
* Fixed a pre-existing malformed AsciiDoc marker in the adjacent `block_expand` entry (`+Default value:*` -> `*Default value:*`) so its default value renders in bold like every other parameter.

All parameter descriptions and defaults were validated against the TinyMCE core formatter source.

---

### **Pre-checks:**

- [x] Branch is correctly prefixed: `hotfix/8/TINYDOC-3531`

- [x] Files have been included where required (partial included by `content-formatting.adoc`).

---

### **Review:**

- [x] Documentation Team Lead has reviewed.

### DIFF
--- a/modules/ROOT/partials/configuration/formats.adoc
+++ b/modules/ROOT/partials/configuration/formats.adoc
@@ -358,7 +358,7 @@ tinymce.init({
 [[expand]]
 ==== `+expand+`
 
-Controls whether the selection is expanded to include the entire matching element before a format is applied or removed. This applies to `+selector+` formats that are not also `+inline+` formats. When enabled, if the selection only partially covers a matching element, the range is expanded to cover the whole element. Setting `+expand+` to `+false+` restricts the operation to the current selection.
+Controls whether the selection is expanded to the enclosing matching selector element before a format is applied or removed. This applies to `+selector+` formats that are not also `+inline+` formats. When enabled, if the selection only partially covers a matching element, the range is expanded to cover the whole element. Setting `+expand+` to `+false+` restricts the operation to the current selection.
 
 *Type:* `+Boolean+`
 
--- a/modules/ROOT/partials/configuration/formats.adoc
+++ b/modules/ROOT/partials/configuration/formats.adoc
@@ -362,7 +362,7 @@ Controls whether the selection is expanded to include the entire matching elemen
 
 *Type:* `+Boolean+`
 
-*Default value:* `+true+` for `+selector+` formats
+*Default value:* `+true+` for `+selector+` formats that are not also `+inline+` formats
 
 *Possible values:* `+true+`, `+false+`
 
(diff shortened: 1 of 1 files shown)


modules/ROOT/partials/configuration/formats.adoc
```diff
@@ -355,6 +355,58 @@ tinymce.init({
...
 ----
 
+[[expand]]
+==== `+expand+`
+
+Controls whether the selection is expanded to include the entire matching element before a format is applied or removed. This applies to `+selector+` formats that are not also `+inline+` formats. When enabled, if the selection only partially covers a matching element, the range is expanded to cover the whole element. Setting `+expand+` to `+false+` restricts the operation to the current selection.
+
+*Type:* `+Boolean+`
+
+*Default value:* `+true+` for `+selector+` formats
```
ArvinJ-H:
```suggestion
*Default value:* `+true+` for `+selector+` formats that are not also `+inline+` formats
```

modules/ROOT/partials/configuration/formats.adoc
```diff
@@ -355,6 +355,58 @@ tinymce.init({
 });
 ----
 
+[[expand]]
+==== `+expand+`
+
+Controls whether the selection is expanded to include the entire matching element before a format is applied or removed. This applies to `+selector+` formats that are not also `+inline+` formats. When enabled, if the selection only partially covers a matching element, the range is expanded to cover the whole element. Setting `+expand+` to `+false+` restricts the operation to the current selection.
```
ArvinJ-H:
```suggestion
Controls whether the selection is expanded to the enclosing matching selector element before a format is applied or removed. This applies to `+selector+` formats that are not also `+inline+` formats. When enabled, if the selection only partially covers a matching element, the range is expanded to cover the whole element. Setting `+expand+` to `+false+` restricts the operation to the current selection.
```

ArvinJ-H (approved):

ShiridiGandham (approved):

ltrouton (approved):

ltrouton (approved):

PR #4276, opened 2026-07-14
TINYDOC-3514 - Add scheduled link checking to CI

kemister85:
**Ticket:** TINYDOC-3514

### What this does

Adds a weekly scheduled link check (Mondays 06:00 UTC, plus manual dispatch). Builds the full production site — all four published versions — and runs [lychee](https://github.com/lycheeverse/lychee) over the built HTML.

Findings go to a single rolling `link-rot` issue: the body is refreshed each run, and the issue closes automatically once a run comes back clean.

Layer 1 of 2 — the PR-time check is #4277.

### Validation

- Config parses under lychee 0.24.2.
- Full local build crawled: tuning `.lycheeignore` took it from **446 errors to 9 (7 distinct)**, all genuine rot — `archive.tinymce.com` (×2), `gsuite.google.com`, `dropbox.com/guide/business`, `ietf.org/rfc/bcp/bcp47.txt`, `addons.mozilla.org`, and `mysite.com/tinymce.html` (a placeholder rendered as a live link, so it needs a content fix rather than an ignore entry). Content follow-up, not blockers here.
- Three hosts confirmed reachable in a browser but refusing automated checkers are now ignored: `stackoverflow.com` (403), `graylog.org` (403), `tinymcedocs.mcp.kapa.ai` (405 — an MCP endpoint, not a page).
- `exit_code` confirmed as a declared output of `lychee-action@v2`, so the reporting conditions fire.
- `link-rot` label created.

**Not yet verified:** the rolling-issue logic cannot run until this is on `main`, because `workflow_dispatch` only fires from the default branch. Dispatch it manually straight after merge to confirm the first real run.

### DIFF
--- a/.github/workflows/link-check.yml
+++ b/.github/workflows/link-check.yml
@@ -1,0 +1,60 @@
+name: Scheduled Link Check
+
+# Weekly crawl of the full published documentation for dead external links.
+# The production playbook (antora-playbook.yml) builds every published version
+# (tinymce/5, /6, /7, /8) into build/site, so this single job covers all
+# versions. Scheduled workflows only run from the default branch, so this file
+# lives on main by design.
+
+on:
+  schedule:
+    - cron: '0 6 * * 1'   # every Monday 06:00 UTC
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  check-links:
+    name: Check external links
+    runs-on: ubuntu-latest
+
+    defaults:
+      run:
+        shell: bash
+
+    steps:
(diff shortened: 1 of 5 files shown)


kemister85:
Closing for now. PR-time link checking in #4277 covers the need, and a weekly full-site scan is not required alongside it. The `.lychee.toml` and `.lycheeignore` configuration this PR introduced is carried by #4277, so nothing is lost. Reopen if a scheduled full-site scan is wanted later — the rolling-issue logic here is untested, since `workflow_dispatch` only fires from the default branch.

PR #4282, opened 2026-07-17
TINYDOC-3432 - Document Vue 3 prop-driven content updates in the Vue technical reference

kemister85:
**Ticket:** TINYDOC-3432

**Site:** `https://pr-4282.tinymce-docs.iad.staging.tiny.cloud/docs/tinymce/latest/vue-ref/#updating-the-editor-content-from-a-parent-component`

**Changes:**

Extends the `v-model` section of the Vue.js technical reference (`modules/ROOT/partials/integrations/vue-tech-ref.adoc`, included by `vue-ref.adoc`) to answer a recurring Kapa question: how the `@tinymce/tinymce-vue` editor updates when a parent component changes the bound content prop after mount.

This brings the Vue reference to parity with the React, Angular, and Svelte references, which already document how each wrapper reacts to parent-driven content updates.

Added a new "Updating the editor content from a parent component" subsection (plus a TOC entry) documenting, verified against `@tinymce/tinymce-vue` v6.3.1 source:

- The component watches the `v-model`/`modelValue` value and calls `editor.setContent()` when a parent reassigns it.
- The guard conditions that prevent a redundant `setContent()` and cursor reset during the normal edit → `update:modelValue` round trip (value is a string, differs from the previous value, and differs from the current editor content compared using `output-format`).
- A `v-model` code example showing a parent reassigning the bound value.
- A single sentence noting that without `v-model` the watcher is inactive, cross-referencing the existing Event binding section for obtaining the editor instance (rather than duplicating that guidance or promoting a binding bypass).

### DIFF
--- a/modules/ROOT/partials/integrations/vue-tech-ref.adoc
+++ b/modules/ROOT/partials/integrations/vue-tech-ref.adoc
@@ -19,6 +19,7 @@
 ** xref:toolbar[`+toolbar+`]
 ** xref:tinymce-script-src[`+tinymce-script-src+`]
 * xref:form-input-bindings-v-model[Form Input Bindings: `+v-model+`]
+** xref:updating-the-editor-content-from-a-parent-component[Updating the editor content from a parent component]
 * xref:event-binding[Event binding]
 
 [[installing-the-tinymce-vuejs-integration-using-npm]]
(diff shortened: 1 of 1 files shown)


kemister85:
@tiny-ben-tran If you can review for next week

PR #4291, opened 2026-07-24
TINYDOC-3562 - Document beforeSetContent listener for AI-only revision history

kemister85:
**Ticket:** TINYDOC-3562

**Site:** [Staging branch](https://pr-4291.tinymce-docs.iad.staging.tiny.cloud/docs/tinymce/latest/revisionhistory/#attributing-ai-assisted-revisions)

**Changes:**
* Document that the TinyMCE AI plugin sets `ai` to `true` on both the `BeforeSetContent` and `SetContent` events, so either event can signal AI involvement.
* Explain both events up front in the "Attributing AI-assisted revisions" introduction, immediately after the note that the Revision History plugin does not set the marker automatically, rather than introducing `BeforeSetContent` later as an addition.
* Present the two events as equally valid choices: `SetContent` alone records the AI-assisted result, and listening to both also preserves the content that preceded each AI edit as its own revision. The choice is left to the integrator's use case.
* Add a `BeforeSetContent` example alongside the existing `SetContent` example in "Saving an AI-assisted revision", showing how to capture the pre-AI content via `editor.getContent()`.

### **Pre-checks:**

- [x] Branch is correctly prefixed: `hotfix/8/TINYDOC-3562`
- [x] Files have been included where required (if applicable).

Prepared for documentation team review.

### DIFF
--- a/modules/ROOT/pages/revisionhistory.adoc
+++ b/modules/ROOT/pages/revisionhistory.adoc
@@ -116,9 +116,14 @@ The metadata is an `+Object+` that holds additional information about a revision
 
 A revision is identified as AI-assisted through the `+source+` field of its xref:#metadata[`+metadata+`] object. When `+metadata.source+` is `+'ai'+`, an AI badge appears on that revision card by default. Set xref:revisionhistory_ai_attribution[`+revisionhistory_ai_attribution+`] to `+false+` to hide the badge.
 
-The {pluginname} plugin does not set this marker automatically. The xref:tinymceai.adoc[{productname} AI] plugin sets `+ai+` to `+true+` on the `+SetContent+` event for any content it inserts. Detecting that signal and recording it as `+metadata.source+` when saving a revision is the responsibility of the application. There are three steps to keep attribution accurate:
+The {pluginname} plugin does not set this marker automatically. The xref:tinymceai.adoc[{productname} AI] plugin sets `+ai+` to `+true+` on both the `+BeforeSetContent+` and `+SetContent+` events for any content it inserts, so either event can signal AI involvement:
 
-. **Detect**: Listen for the `+SetContent+` event and record when its `+ai+` property is `+true+`.
+* `+BeforeSetContent+` fires immediately before the AI content is inserted. Calling `+editor.getContent()+` at this point returns the content as it stood before the AI edit.
+* `+SetContent+` fires immediately after the AI content is inserted. Calling `+editor.getContent()+` at this point returns the AI-generated result.
+
+Either event can drive attribution, and both can be used together. Listening for `+SetContent+` alone records the AI-assisted result; listening for both also preserves the content that preceded each AI edit as its own revision. Detecting the signal and recording it as `+metadata.source+` when saving a revision is the responsibility of the application. There are three steps to keep attribution accurate:
+
+. **Detect**: Listen for `+BeforeSetContent+`, `+SetContent+`, or both, and record when the `+ai+` property is `+true+`.
 . **Save**: When saving a revision, set `+metadata.source+` to `+'ai'+` if AI contributed since the last save.
 . **Restore**: When a revision is restored, carry its marker forward so later revisions remain correctly attributed.
 
@@ -155,14 +160,7 @@ const saveRevision = () => {
 };
 ----
 
-==== Capturing content before and after an AI edit
-
-The previous example marks the next saved revision as AI-assisted, but that revision can combine human edits made since the last save with the AI-generated change. To keep the two apart, save a revision on the `+BeforeSetContent+` event as well. Both events carry an `+ai+` property that is `+true+` when the xref:tinymceai.adoc[{productname} AI] plugin inserts content:
-
-* `+BeforeSetContent+` fires immediately before the AI content is inserted. Calling `+editor.getContent()+` at this point captures the latest non-AI content, which can be saved as a revision that reflects the human-authored state.
-* `+SetContent+` fires immediately after the AI content is inserted. Calling `+editor.getContent()+` at this point captures the AI-generated result, which can be saved as an AI-assisted revision.
-
-Listening for both events produces a finer-grained history that distinguishes purely AI-generated content from the human content that preceded it:
+A revision saved this way can combine human edits made since the last save with the AI-generated change. To keep the two apart, save a revision from `+BeforeSetContent+` as well. This records the content that preceded each AI edit as its own revision, producing a finer-grained history:
 
 [source,js]
 ----
@@ -191,8 +189,6 @@ editor.on('SetContent', (e) => {
   }
 });
 ----
-
-Listening for `+BeforeSetContent+` is optional. Use `+SetContent+` alone to record only the AI-assisted result, or pair it with `+BeforeSetContent+` when the revision history should also preserve the content as it was before each AI edit. Choose the approach that suits the application's use case.
 
 === Restoring an AI-assisted revision
 
--- a/modules/ROOT/pages/revisionhistory.adoc
+++ b/modules/ROOT/pages/revisionhistory.adoc
@@ -155,6 +155,45 @@ const saveRevision = () => {
 };
 ----
 
+==== Capturing content before and after an AI edit
+
+The previous example marks the next saved revision as AI-assisted, but that revision can combine human edits made since the last save with the AI-generated change. To keep the two apart, save a revision on the `+BeforeSetContent+` event as well. Both events carry an `+ai+` property that is `+true+` when the xref:tinymceai.adoc[{productname} AI] plugin inserts content:
+
+* `+BeforeSetContent+` fires immediately before the AI content is inserted. Calling `+editor.getContent()+` at this point captures the latest non-AI content, which can be saved as a revision that reflects the human-authored state.
+* `+SetContent+` fires immediately after the AI content is inserted. Calling `+editor.getContent()+` at this point captures the AI-generated result, which can be saved as an AI-assisted revision.
+
+Listening for both events produces a finer-grained history that distinguishes purely AI-generated content from the human content that preceded it:
+
+[source,js]
+----
+// Save the pre-AI (human) content before the TinyMCE AI plugin inserts its edit.
+editor.on('BeforeSetContent', (e) => {
+  if (e.ai) {
+    const revision = {
+      revisionId: createRevisionId(), // Replace with your ID generation
+      createdAt: new Date().toISOString(),
+      content: editor.getContent()
+    };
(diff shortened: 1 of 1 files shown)


modules/ROOT/pages/revisionhistory.adoc
```diff
@@ -155,6 +155,45 @@ const saveRevision = () => {
...
+      createdAt: new Date().toISOString(),
+      content: editor.getContent(),
+      metadata: { source: 'ai' }
+    };
+    saveToStorage(revision); // Replace with your storage call
+  }
+});
+----
+
+Listening for `+BeforeSetContent+` is optional. Use `+SetContent+` alone to record only the AI-assisted result, or pair it with `+BeforeSetContent+` when the revision history should also preserve the content as it was before each AI edit. Choose the approach that suits the application's use case.
```
metricjs: "optional" here implies that listening to setcontent isn't. Whereas the correct framing is that integrators can use either or both, we don't care - it's entirely up to their use case

modules/ROOT/pages/revisionhistory.adoc
```diff
@@ -116,9 +116,14 @@ The metadata is an `+Object+` that holds additional information about a revision
...
+The {pluginname} plugin does not set this marker automatically. The xref:tinymceai.adoc[{productname} AI] plugin sets `+ai+` to `+true+` on both the `+BeforeSetContent+` and `+SetContent+` events for any content it inserts, so either event can signal AI involvement:
 
-. **Detect**: Listen for the `+SetContent+` event and record when its `+ai+` property is `+true+`.
+* `+BeforeSetContent+` fires immediately before the AI content is inserted. Calling `+editor.getContent()+` at this point returns the content as it stood before the AI edit.
+* `+SetContent+` fires immediately after the AI content is inserted. Calling `+editor.getContent()+` at this point returns the AI-generated result.
+
+Either event can drive attribution, and both can be used together. Listening for `+SetContent+` alone records the AI-assisted result; listening for both also preserves the content that preceded each AI edit as its own revision. Detecting the signal and recording it as `+metadata.source+` when saving a revision is the responsibility of the application. There are three steps to keep attribution accurate:
```
metricjs: "Listening for `+SetContent+` alone records the AI-assisted result; listening for both also preserves the content that preceded each AI edit as its own revision." doesn't quite read right to me. 

Using one event (doesn't matter which) means you could get revisions that mix AI and human generated content, since it's just "changes since last revision" - it just depends which order. E.g.:

1. Listen only for beforesetcontent = revision could be human changes + a previous AI change
2. Listen only for setcontent = revision could be AI changes + previous human changes.

If integrators specifically want to differentiate AI and human changes, we'd recommend listening to both to get the right granularity of data. If they just want to save a revision because changes are being made, they can pick whichever event suits their use cases better.

modules/ROOT/pages/revisionhistory.adoc
```diff
@@ -116,9 +116,14 @@ The metadata is an `+Object+` that holds additional information about a revision
...
+
+. **Detect**: Listen for `+BeforeSetContent+`, `+SetContent+`, or both, and record when the `+ai+` property is `+true+`.
```
metricjs: I wouldn't say "record" that the ai property is true - they don't need to write that down. They want to listen for beforesetcontent and/or setcontent and **check** that the ai property is true, and only then do they save a revision (for the purposes of these instructions - obviously they can save revisions at other points too). Otherwise they'd be saving revisions with, like, every word that's typed or something - can't quite remember how often setcontent runs but iirc it's _often_

modules/ROOT/pages/revisionhistory.adoc
metricjs: This... would work, but I'd make it simpler and just call `saveRevision()` here instead of having a global state thing? Probably pass in an 'ai' param like so: `saveRevision(true)` which sets the `ai` property, so the function can be reused for non-ai saves.

modules/ROOT/pages/revisionhistory.adoc
```diff
@@ -155,6 +160,36 @@ const saveRevision = () => {
 };
```
metricjs:
```suggestion
const saveRevision = (aiAssisted) => {
 const revision = {
   revisionId: createRevisionId(), // Replace with your ID generation
   createdAt: new Date().toISOString(),
   content: editor.getContent(),
   // Mark the revision as AI-assisted if AI contributed since the last save.
   ...(aiAssisted ? { metadata: { source: 'ai' } } : {})
 };

 saveToStorage(revision); // Replace with your storage call
};
```
following from comment above. you'd also remove line 139 if doing it this way

modules/ROOT/pages/revisionhistory.adoc
```diff
@@ -155,6 +160,36 @@ const saveRevision = () => {
...
 ----
 
+A revision saved this way can combine human edits made since the last save with the AI-generated change. To keep the two apart, save a revision from `+BeforeSetContent+` as well. This records the content that preceded each AI edit as its own revision, producing a finer-grained history:
+
+[source,js]
```
metricjs: Should've kept reading. This example is more like what I would expect, following from comments above. I'd either use this format above, or for both, do basically this but split the repeated part out to a function.

modules/ROOT/pages/revisionhistory.adoc
```diff
@@ -155,6 +160,36 @@ const saveRevision = () => {
...
+  if (e.ai) {
+    const revision = {
+      revisionId: createRevisionId(), // Replace with your ID generation
+      createdAt: new Date().toISOString(),
+      content: editor.getContent(),
+      metadata: { source: 'ai' }
+    };
+    saveToStorage(revision); // Replace with your storage call
+  }
+});
```
metricjs, line 170: Example, in case you want it:
```suggestion
const saveRevision = (aiAssisted) => {
  const revision = {
      revisionId: createRevisionId(), // Replace with your ID generation
      createdAt: new Date().toISOString(),
      content: editor.getContent(),
      ...(aiAssisted ? { metadata: { source: 'ai' } } : {})
    };
    saveToStorage(revision); // Replace with your storage call
};

editor.on('BeforeSetContent', (e) => {
  if (e.ai) {
    saveRevision(false);
  }
});

// Save the post-AI content and mark it as AI-assisted.
editor.on('SetContent', (e) => {
  if (e.ai) {
    saveRevision(true);
  }
});
```

metricjs (approved):
A couple details need clarification, and the examples aren't consistent - left comments. But almost there

modules/ROOT/pages/revisionhistory.adoc
kemister85: Agreed. "optional" is gone, neither event is framed as the default now. The choice is presented purely in terms of the granularity the integrator needs, with either/both equally valid.

modules/ROOT/pages/revisionhistory.adoc
kemister85: Reworked to your framing

- BeforeSetContent alone: human edits + the AI edit that preceded them
- SetContent alone: the AI edit + the human edits that preceded it
- both: the two separated into distinct revisions

Closing line makes the recommendation conditional rather than absolute, either event is fine if the goal is just "save a revision when content changes".

modules/ROOT/pages/revisionhistory.adoc
kemister85: Changed to "check whether the ai property is true", with the save happening only when that check passes.

One correction on the reasoning though: `SetContent` doesn't fire per keystroke. It's dispatched from `editor.setContent()`, `editor.insertContent()` and `selection.setContent()`(core/content/PrePostProcess.ts), so paste, image insert, and similar — not typing. The guard is still needed for exactly the reason you give, so I've kept the point but worded it as "both events also fire for content changes that the AI plugin did not make" rather than tying it to typing.

Does this work for you?

modules/ROOT/pages/revisionhistory.adoc
kemister85: Done. `saveRevision` now takes an `aiAssisted` parameter and the module-level flag is gone entirely.

modules/ROOT/pages/revisionhistory.adoc
kemister85: Applied, and the old line 139 flag is removed as you noted.

modules/ROOT/pages/revisionhistory.adoc
kemister85: Went with the split-into-a-function option. The helper is defined once up front, then the single-event and both-events examples are just the listeners calling it. Let me know if this is fine?

modules/ROOT/pages/revisionhistory.adoc
kemister85: Used this, with the helper hoisted out above so the single-event example can share it
rather than defining it twice.

modules/ROOT/pages/revisionhistory.adoc
```diff
@@ -116,55 +116,87 @@ The metadata is an `+Object+` that holds additional information about a revision
 
 A revision is identified as AI-assisted through the `+source+` field of its xref:#metadata[`+metadata+`] object. When `+metadata.source+` is `+'ai'+`, an AI badge appears on that revision card by default. Set xref:revisionhistory_ai_attribution[`+revisionhistory_ai_attribution+`] to `+false+` to hide the badge.
 
-The {pluginname} plugin does not set this marker automatically. The xref:tinymceai.adoc[{productname} AI] plugin sets `+ai+` to `+true+` on the `+SetContent+` event for any content it inserts. Detecting that signal and recording it as `+metadata.source+` when saving a revision is the responsibility of the application. There are three steps to keep attribution accurate:
+The {pluginname} plugin does not set this marker automatically. The xref:tinymceai.adoc[{productname} AI] plugin sets `+ai+` to `+true+` on both the `+BeforeSetContent+` and `+SetContent+` events for any content it inserts, so either event can signal AI involvement:
```
metricjs: In all our edits, we've lost any early mention of _why_ you should listen for these events. 

```suggestion
The {pluginname} plugin does not set this marker automatically. Integrators should listen for the relevant content events, and save revisions manually. The xref:tinymceai.adoc[{productname} AI] plugin sets `+ai+` to `+true+` on both the `+BeforeSetContent+` and `+SetContent+` events for any content it inserts, so either event can signal AI involvement:
```
Not the best wording, but something along these lines?

metricjs:
Screenshot 2026-07-30 at 11 51 12 am
Do we still need both these examples really? I expected either one example with both events, or one example for each event. This structure aligns more with the old structure of the section, with the initial focus on just setcontent

tiny-ben-tran (approved):

PR #4317, opened 2026-08-20
DO NOT MERGE - TINYDOC-3514 link check verification

kemister85:
Throwaway pull request that exercises the automated external-link check in #4277. **Not for merge — close once the run is confirmed.**

Base is the #4277 head so the new workflow is the one that runs.

Expected outcome in the job summary:

| Link | Expected |
|------|----------|
| `developer.mozilla.org/.../TinyDoc3514DoesNotExist` (prose) | reported broken (404) |
| `w3.org/TR/wai-aria-1.2/` | passes |
| `developer.mozilla.org/.../Window/postMessage` | passes |
| `xref:events.adoc` | not checked |
| `www.tiny.cloud/docs/...` | excluded |
| `npmjs.com/package/tinymce` | ignored |
| all six URLs inside the `[source,js]` block | never reach the checker |

The last row is the point of the exercise: the previous extractor read raw AsciiDoc and would have checked all six.

### DIFF
--- a/.github/workflows/preview_create.yml
+++ b/.github/workflows/preview_create.yml
@@ -96,13 +96,15 @@ jobs:
       id: lychee
       uses: lycheeverse/lychee-action@v2
       with:
-        args: --config .lychee.toml --no-progress --format json './build/docs/tinymce/latest/**/*.html'
+        args: --config .lychee.toml --no-progress './build/docs/tinymce/latest/**/*.html'
+        format: json
         output: lychee/results.json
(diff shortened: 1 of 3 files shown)


kemister85:
Verification complete: the built-HTML link check reported 1 of the 2 dead links added — the prose anchor — and correctly ignored the code-block twin plus the five other code-sample URLs. Evidence is recorded in #4277. Closing; this page was never for merge.

PR #4352, opened 2026-09-21
TINYDOC-3598: The `diff` API would show identical tables as a replacement due to whitespace in the input strings.

kemister85:
**Ticket:** TINYDOC-3598 (TINYMCE-14568)

**Site:** [Staging](https://pr-4352.tinymce-docs.iad.staging.tiny.cloud/docs/tinymce/latest/8.9.2-release-notes/#the-diff-api-would-show-identical-tables-as-a-replacement-due-to-whitespace-in-the-input-strings)

**Changes:**
* Added release note entry for TINYMCE-14568 to `modules/ROOT/pages/8.9.2-release-notes.adoc` (Accompanying Premium plugin changes section, **Revision History**): the `diff` API would show identical tables as a replacement due to whitespace in the input strings.

### **Pre-checks:**

- [x] ~~Branch is correctly prefixed~~ (release-note branch)
- [x] Files have been included where required (if applicable).

---

### **Review:**

- [x] Documentation Team Lead has reviewed.


modules/ROOT/pages/8.9.2-release-notes.adoc
```diff
@@ -72,6 +72,22 @@ The {productname} {release-version} release includes an accompanying release of
...
+=== Revision History
+
+The {productname} {release-version} release includes an accompanying release of the **Revision History** premium plugin.
+
+**Revision History** includes the following fix.
+
+==== The `+diff+` API would show identical tables as a replacement due to whitespace in the input strings
+// #TINYMCE-14568
+
+Previously, the xref:revisionhistory.adoc#diff[`+diff+`] API compared the two HTML strings exactly as supplied, so the whitespace used to format them registered as content. Two fragments that rendered identically, such as the same table indented differently, were reported as a replacement instead of as unchanged content.
```
MitchC1999: I think the whitespace comparison resulting in a diff was more of a bug in visual dom diff, not that the whitespace was treated as content.

kemister85: Good catch, thanks and yes you're right, the release note shouldn't be making that claim either way. I've dropped the cause from both paragraphs so the entry describes only the observable behavior

modules/ROOT/pages/8.9.2-release-notes.adoc
```diff
@@ -71,6 +71,20 @@ The {productname} {release-version} release includes an accompanying release of
...
+=== Revision History
+
+The {productname} {release-version} release includes an accompanying release of the **Revision History** premium plugin.
+
+**Revision History** includes the following fix.
+
+==== The `+diff+` API would show identical tables as a replacement due to whitespace in the input strings
+// #TINYMCE-14568
+
+Previously, the xref:revisionhistory.adoc#diff[`+diff+`] API compared the two HTML strings exactly as supplied. Two fragments that rendered identically, such as the same table indented differently, were reported as a replacement instead of as unchanged content.
```
MitchC1999:
```suggestion
Previously, the xref:revisionhistory.adoc#diff[`+diff+`] API compared the two HTML strings exactly as supplied. When the strings contained whitespace characters, for example, indentation, the API could produce an unexpected result. Specifically, identical strings containing an indented table in HTML were reported as a replacement instead of as unchanged content.
```

MitchC1999 (approved):

tiny-ben-tran (approved):

